PR #3, opened 2020-09-08
Add Pre-Commit hook and .gitignore for idea files

holtzmak:


MummenRider (approved):
Pre-commit hook and .gitignore looks good!

Issue #33, opened 2020-10-16
Investigate one more multiplatform project template


holtzmak:
After investigating Kotlin's Multiplatform Mobile (KMM) project template in #26, it would be in our best interest to try out at least one more way of creating a multiplatform project to see if it's easier to use.

(Copied from #26, try out something similar to the following):

Investigate how to write/get working the cross-platform (agnostic) GUI code:
1. Start investigating another, well-known multiplatform project/language
2. Could start a Wiki page with the options cross-platform code could be written
3. Get a window running on android and iOS (this is expected to be a dummy window, if you write the "real-thing" we could also merge into our codebase!)

MummenRider:
My experience : Setting up flutter in macOS is fairly easy. I haven't tried this on windows yet.
Future Goal: Explore Flutter documentation and maybe see Dart Language

holtzmak:
> My experience : Setting up flutter in macOS is fairly easy. I haven't tried this on windows yet.
> Future Goal: Explore Flutter documentation and maybe see Dart Language

Does it feel like Flutter is strong competition for KMM with its better integration with Android Studio? If we all end up using the same IDE (especially one that can run iOS without XCode* specifically) we could probably put configuration files under revision control so people don't have to figure out on their own (do it once kinda thing).

*I'm pretty sure you still had to download XCode tools, but that's minimal compared to the 40GB XCode can inflate to

mansiPaTeLL:
@MummenRider Great. Thanks to look into this. I will also help you out here with the investigations.

MummenRider:
If you want to use ios simulator, then downloading xcode is required. Flutter is indeed a huge competitor to KMM and we should really consider using flutter instead. As for IDE, we don't need huge IDE such as intellij, android studio or xcode, we can use a lightweight ide like VScode. I am closing this as I believe that I investigated enough to resolve this issue.

PR #35, opened 2020-10-21
Import Kotlin Multiplatform Mobile project and run pre-commit hooks

### LINKED ISSUE #26: Investigate (and create) cross-platform agnostic GUI code
Investigate how to write/get working the cross-platform (agnostic) GUI code:
1. Start investigating Kotlin's cross-platforming https://kotlinlang.org/lp/mobile/ 
1. Could start a Wiki page with the options cross-platform code could be written
1. Get a window running on android and iOS (this is expected to be a dummy window, if you write the "real-thing" we could also merge into our codebase!)
1. Investigate different GUI libraries for cross-platform Kotlin (if we aren't constrained by the API)*

*Kotlin works with all Java code, so we can investigate Java GUI libraries like JavaFX and Java Swing. To avoid overhead of JavaFX we could use TornadoFx (Kotlin wrapper library) instead, for example.
 
See the above notes to determine how to move forward.

holtzmak:
Closes #26 once complete and tested on windows, or will directly close regardless of review if #33 goes ahead instead.

**If we want to really go with KMM**, still need to do:

- [ ] Wiki page with IDE config instructions (will be a little lengthy, but should be a one-time configuration)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,106 @@
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
-/.idea/
+
+# IntelliJ files, caches, build files, git-related etc.
+*.iml
+.gradle/
+build/
+.idea/
+.DS_STORE
+local.properties
+.cache/
+
+# Xcode
+#
+# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
+
+## User settings
+xcuserdata/
+
+## compatibility with Xcode 8 and earlier (ignoring not required starting Xcode 9)
+*.xcscmblueprint
+*.xccheckout
+
+## compatibility with Xcode 3 and earlier (ignoring not required starting Xcode 4)
+build/
+DerivedData/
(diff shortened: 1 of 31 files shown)


holtzmak:
@MummenRider lemme know if you want a walk-through review on this one, typically we wouldn't make a PR this huge, but there's no way around it for the project base.

holtzmak:
Further notes:
1. This runs with JDK 14, SDK Android 29, Gradle 6.6.1 for building, and older gradle tools 3.5.2 (because we use AndroidX which relies on this)
2. This has been tested on mac so far with the Google Pixel 3A ("Hello World") and iPhone 11 (No base app, but project compiles and an empty app icon appears due to `dummyFile`)

 _I've personally not worked with Swift before so I made no "Hello World" app in Swift, but can do so if we're moving forward with KMM_

MummenRider:
LGTM

MummenRider (approved):
LGTM

holtzmak:
Closing this PR as per #33 - we'll be using Flutter, which is similar but with better support. Uses the dart language over kotlin.

PR #58, opened 2020-11-15
Add the Transfer of Care UJMs

holtzmak:
A first draft of the Transfer of Care stories, in UJM form to discover landing pages in the application and what the transporter needs from the pages. Close #51 .

PR #60, opened 2020-11-15
added ATR ujms

### LINKED ISSUE #48: Create the Animal Transport Record UJMs from USMs
Using the USM from #41, create the corresponding User Journey Maps (UJMs) using the structure from #47.

SanaKhan1997:
Closes #48


holtzmak (changes requested):
I see what you were referring to in [this PR comment](https://github.com/holtzmak/Humane-Transport/pull/58#issuecomment-727559442). I like what you have done. 

I'm thinking about how much different the UJMs are from the USMs because they convey the same information, and if we actually need the UJMs. But, I think we still have use for the UJMs with respect to my first draft of ToC UJMs - we can use them to determine where the responsibilities of phases lie in our app pages.

I have a request to change your commits though, **it's not good etiquette to merge into someone else's branch (especially if it's in progress)**. You can see that I'm rearranging the `documentation` folder how I see fit - what you should do is put your UJMs in `documentation` as everything else already is (meaning, you should be merging into `master`). What will then happen is that I will have to rebase on `master` before merging, and will address the changes made by your PR. This method of doing only your changes will result in less work (and risk of messing history up) later.
- [ ] Change your commits to merge into `master` and not my branch. Let me know if you need help doing this.

SanaKhan1997:
@holtzmak Thanks for pointing out the base branch issue. I have changed the base branch to master hope that will fix the issue and let me know if its good to merge now.

holtzmak:
> @holtzmak Thanks for pointing out the base branch issue. I have changed the base branch to master hope that will fix the issue and let me know if its good to merge now.

Note quite, you still have my commits (which we definitely don't want on `master` now). What you want is your commit `c0f059aeecdacba73ac74541b689716a05bd4438` on top of what's in `master` right now.

SanaKhan1997:
@holtzmak needs approval for merging

holtzmak (approved):
LGTM

PR #62, opened 2020-11-16
FWR UJM added

MummenRider:
Here's a UJM for FWR using Figma. If team decided to use the first template, I will change this for consistency.


holtzmak:
> Here's a UJM for FWR using Figma. If team decided to use the first template, I will change this for consistency.

Ok, will host a poll in the group chat for this before reviewing

mansiPaTeLL:
@MummenRider I think using Figma UJM, it describe the user flow/feeling really well, which is required for prototyping. I think this is also really good. But, we also need to have "Doing" "Thinking", "Saying" section to describe the journey well.

holtzmak:
@MummenRider Is this the only FWR UJM to make?

SanaKhan1997:
You created a UJM for a perfect FWR scenario but what if a transporter notice an animal that requires FWR before its next interval due time(in that case shortest FWR interval will be recorded for all as this record does not apply to individual animal instead it provides overview). This case was discussed in last client meeting.

MummenRider:
> @MummenRider I think using Figma UJM, it describe the user flow/feeling really well, which is required for prototyping. I think this is also really good. But, we also need to have "Doing" "Thinking", "Saying" section to describe the journey well.
@mansiPaTeLL 
The action and doing is included in the template. For example: 1. Opening the app.
For Thinking and saying, the comment box is the representation of that

MummenRider:
> @MummenRider Is this the only FWR UJM to make?
@holtzmak 
No, I have 3 more to do once we decided which template we're going to use. 
I don't want to commit to this template yet and do all the UJMs there is a no guarantee yet that we're going to use the template.
I created this 1 UJM for the template.  If we decided to use Figma style, then I will create the rest

MummenRider:
> You created a UJM for a perfect FWR scenario but what if a transporter notice an animal that requires FWR before its next interval due time(in that case shortest FWR interval will be recorded for all as this record does not apply to individual animal instead it provides overview). This case was discussed in last client meeting.

@SanaKhan1997 , I do have a different UJM for this scenario, I did 1 perfect scenario only to show the template for Figma.
If we decided to use this style, then I'll do the rest.

holtzmak:
> > @MummenRider Is this the only FWR UJM to make?

> @holtzmak
> No, I have 3 more to do once we decided which template we're going to use.
> I don't want to commit to this template yet and do all the UJMs there is a no guarantee yet that we're going to use the template.
> I created this 1 UJM for the template.  If we decided to use Figma style, then I will create the rest

As per the poll in the group chat, using both templates (Figma and NN/g) at the same time is fine.

holtzmak (approved):
Your figma template is very similar to the NN/g template, which is great for when we generate lofis. A difference is that your figma template is somewhat application focused - but I think our application will very much so reflect the sequence of events in transport: making it a moot point to argue whether application details should be in the UJMs or not.

Do make note to mark outstanding FWR occurrences in the UJM, even if copy-pasted from the USMs. Looking forward to the rest of the FWR UJMs.

holtzmak (changes requested):
Changes still approved, some minor changes I'd like to see. 

1. The [exceeded MTI form](https://github.com/holtzmak/Humane-Transport/blob/26cdf57fca78b17e8f3644650ecc5d1a686ae070/MTIExceededScenarioUJM.pdf) was really well done, good job.
2. Even though your UJMs mention application interaction, they easily apply to the paper form interaction so no changes there.
3. Really digging the new emotion line with the faces, it shows better if a situation is good or not for the transporter (like feeling poorly if some event has occurred that puts the situation in jeopardy).

I'd like to see one change (and you're free to do what you want with the second):
- [ ] Please put the UJMs into a folder named `feed, water, rest ujms` in the `documentation` folder in the codebase like the other UJMs are now.

And, I've seen the word "thinking" misspelled "thingking" in several forms, might want to fix that if you want to bother.

PR #64, opened 2020-11-16
Add Contingency Plan UJMs

### LINKED ISSUE #50: Create the Contingency Plan UJMs from USMs
Using the USM from #43, create the corresponding User Journey Maps (UJMs) using the structure from #47.

holtzmak:
Closes #50


SanaKhan1997 (approved):
Thanks! I really like the details of injured or perished animal scenario.

PR #65, opened 2020-11-17
Added CFIA Inspection-related-ujm

### LINKED ISSUE #52: Create the CFIA Inspector-related (and other) UJMs from USMs
Using the USM from #45, create the corresponding User Journey Maps (UJMs) using the structure from #47.

mansiPaTeLL:
Closes #52


holtzmak (changes requested):
- [x] Please change the folder name to lowercase as per convention

_We usually have things lowercase in codebases as per language convention (usually class files are camel case for example)_

holtzmak (commented):
Some insights I had while reviewing (some of your insights are related):
1. I'm not sure if the application should be used as a file system for everything related to the transport. That's the future idea, that our application is a step towards, but if one application takes on too much responsibility, it will grow too large and unwieldy and will eventually stop being designed for its users. **Meaning, CFIA inspector orders probably don't need to be copied into app, but we should note that there were orders**
2. I think if the CFIA orders some change to transport route, animals onboard, driver change, etc. **we will need to provide transporters a "page" to make note of this**, but like the email feature for CFIA agents to get the forms off app, the app should only do transporter work.
3. I think if the transport is 6+ hours, and the transporter does not notify vet, that the transporter has done something wrong and CFIA will have to catch this when reviewing the transport docs and the transporter will have to defend themselves because they didn't do what they were supposed to (I think the app shouldn't need to remind them of their responsibilities in the way you suggest)

SanaKhan1997:
I agree with @holtzmak. Even though some points suggested in UJMs can be good but they are not expressing current user journey but instead they represent what could be done through this application which might make our prototype stage more complex. Remember at this stage our app is only focused on transportation process and inspection related task can be as simple as "retrieve information"(Driver's training license, transport forms) to show inspector.

But some of the suggestion you made through UJMs are really good. Thank you!

holtzmak (commented):
Looks like you have 6 files (duplicates?) somehow, could you fix this?

holtzmak (approved):
LGTM

PR #72, opened 2020-11-22
lo-fi landing pages

MummenRider:
After log in page - 5 Different designs
Form page - 2 different designs
Log in page - 1 design
Side bar - 1 design


mansiPaTeLL:
Thanks @MummenRider. As Clark mentioned above, we have generated different design for each pages. Please let us know if you have any suggestions on our design. @holtzmak @SanaKhan1997

holtzmak:
Will have more comments in the future, while I'm here now could someone please:
- [ ] make a `gui prototypes` folder in`documentation`, and in that we can make another `lofi` folder hold these! 

Since we can always shoehorn in more features in the future we should date this batch of lofis for future reference.

holtzmak (approved):
Lovely first round of lofis @MummenRider and @mansiPaTeLL .
1. After login page/home page:
I really like design # 4. I haven't really needed to see my profile right away after opening an application that I use pretty much daily. It also hides transportation information from onlookers when first opening the app (who knows who may be watching). An argument for design # 5 can be made because travel history is important.
2. Transporter's profile:
It would be nice if we had a big, shiny, obvious "EDIT" button for transporters. The drop-down menu that's common on most applications often has edit in there but it's not intuitive to me. What purpose does the drop-down menu have in this page if we have most menu options at the bottom of the screen?
3. In-progress ATR page(s):
I can see a combination of the designs working. This particular page has room for us to incrementally work up to its final design because window navigation and saving form elements is something we have to think about while developing. For MVPs I see we first would have one infinitely-scrolling page for the form, then we could have a page to "jump to" sections of the form page, then we break the form into pages (hard to possibly save elements while jumping around), and then add fun features like the progress bar.
4. Login page: It's solid, nothing fancy.
Maybe we could have a little (?) icon that people who inadvertently get their hands on the app can click and have it take them to our website.
5. Sidebar menu page:
I mentioned previously that this design hadn't made sense to me. Seeing it now, I would remove mention of the drivers name and links to profile (that's already at bottom of the screen) and instead make the side menu a helpful link bar that goes to CFIA sites, our website for the application (FAQ!), etc. We will need to be careful developing that if someone opens a link off-app that they don't lose their information on the page they happened to have open.

- [ ] I have some suggestions in the above paragraph (sorry, long read) that would be nice to see addressed. If any questions, let me know.

Otherwise, these are good and given my above input I think are ready for midfis (of course, we will want to see @SanaKhan1997 's input as well).

SanaKhan1997:
Looks great! I do have few concerns though.

1. For Login: Design is simple which is nice but my only question would be to have a link to register as well for someone who's not a user already( I know we talked about having transporters info already registered by administrators but right now we dont have any feature that'll allow staff to do that).
2. After Login: Design 4 is awesome! I know its a lo-fi but I would like to see a little hover text on driver image("edit profile").  Design 1 also keeps it simple as it should be since its the first thing transporter will see once logged in but button for editing profile should also be there.
3. Form Page: I think both designs will work well but one thing as user that would frustrate me is the scrolling part and clicking next/prev to get to desired form section(instead we could have a drop down that allows user to select the part they want to edit). Overall design 2 seems more effiecient.
4. Side bar: Pretty simple and neat for a user as it should be.
5. Drivers Profile: As suggested above, it would be nice to see edit feature in the profile. I think driver training info and license number should be highlighted somehow.

Overall you guys have put so much effort into these Lo-fis. Thanks!

MummenRider:
@holtzmak , @SanaKhan1997 : Thanks for your feedbacks. We'll make sure to implement all of your suggestions in our mid-fi!

holtzmak:
@mansiPaTeLL lmao where's the files?

mansiPaTeLL:
> @mansiPaTeLL lmao where's the files?

@holtzmak I deleted the files as it was not in the folder. I am going to upload the prototype folder and will merge it with the master to make it more organized.

PR #76, opened 2020-12-14
Add project progress vlog 3 links

holtzmak:


holtzmak:
Merged without review as review not required

PR #87, opened 2021-01-05
Add role based access security description

### LINKED ISSUE #68: Investigate (and create) Role-Based Access and Security Provisions for the application
We've since decided that transporters need to login to the application to use it because:
1. Valid login information protects transporters from malicious attempts to submit fake animal transport forms in their name
2. Prevents unauthorized persons from using the application (we don't want or need the general public submitting fake animal transport forms)
3. Reduces database traffic just to persons who need to use it (piggybacking on point 2)

So, we need to investigate how to document the access and security provisions we should have to make the above possible. *I'd imagine this just to be documenting the login process and information collected.*

***Consider also the discussion pervious about now providing a way for transporters to sign up on the application, and that maybe the administration application must create an account on their behalf.*** We might not do this now, as the administration application is separate but we can still think about it.

Some ethical questions we should ponder now are: Do we need to collect information like the transporter's birthday? Do we want to start collected Canadian Livestock Transporter(s) numbers for identification early, even though we can't verify the numbers now? etc. We definitely don't want to collect too much information, what if the database is hacked?

This task is done when:
- [x] We have a method or structure of documenting transporter access and security provisions to our application
- [x] We have designed, or agreed on, what provisions the transporter has to access our application (login details, how to sign up, etc.) **This could be made into a new task, up to you**
- [x] You **MUST** consider how we store login information in the case the database is hacked. Do we hash the information when storing the information? We have to protect transporters using our application just as they have to protect themselves.

holtzmak:
Closes #68.

I've added the descriptions as described in the task. Since Transporter's are our main persona (and the only one we're dealing with until admin application is made (if)), their permissions are fairly "baked in". 

I believe the database entity diagrams will better describe what information we collect, and our encryption schemes are yet to be investigated in depth.


MummenRider (approved):
As discussed during our sprint meeting, everything looks good and ready to be merged

mansiPaTeLL:
It looks good to me. Thanks

Issue #78, opened 2020-12-18
Experiment with Flutter GUI


holtzmak:
Using the project framework set up in #56 , experiment with the following items:
* Make a widget (window)
* Make a sticky navigation bar (a bar that is on all windows/widgets)
* Make a sliding menu (a menu you swipe left/right to show/hide)
* Make our progress bar imitation (think of the circles in the GUI prototypes) (consider how we would update progress)

You're free to play with other GUI elements too, buttons, colors, etc.

Any code you write to do the above can be merged into the codebase - we'll be able to overwrite/copy/etc. when we go to create the final GUI windows/widgets. You may want to keep a list of references that helped you: for example, Flutter cookbook pages - decide if we should have a dedicated wiki page for Flutter cookbook items.

MummenRider:
- [x] Create a Bottom Navigation Bar.  (Note: This task should include creating widgets)
- [x] Create a sliding window
- [ ] Create progress bar (Note: This task may take a lot of time to finish due to its state complexity)

MummenRider:
Creating progress bar would require strong knowledge about state management system which I believe deserve its own issue. I will create a new issue for this in the future.

PR #90, opened 2021-01-07
Added basic implementation of Drawer widget

MummenRider:
Sources:
https://flutter.dev/docs/cookbook/design/drawer
Note: All designs here are only temporary, will change the styling once we finalized our design


app/lib/screens/home/home.dart
```diff
@@ -1,10 +1,12 @@
-import 'package:app/navigation.dart';
+import 'package:app/navbar/navigation.dart';
+import 'package:app/sidebar/sidebar.dart';
 import 'package:flutter/material.dart';
 
 class HomePage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
+      drawer: SideBar(),
```
holtzmak, line 9: I'm wondering, since we'll likely be sharing the same drawer across widgets (windows/pages/different screens) if we shouldn't give the widgets the drawer to use (injection).

Do you know if the `Widget build(injected objects)` runs only once, or is it called every time the `HomePage` is made active?

app/lib/screens/in_progress/in_progress.dart
```diff
@@ -8,9 +8,8 @@ class InProgress extends StatefulWidget {
 class _InProgressState extends State<InProgress> {
   @override
   void initState() {
-    // TODO: implement initState
```
holtzmak, line 11: In the future, we'll make nicer TODOs with reference to tasks, like: `TODO: #78`

app/lib/sidebar/local_widgets/list_tile.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  CustomTile({@required this.icon, @required this.text}) {
+    assert(icon != null);
+    assert(text != null);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Padding(
+      padding: const EdgeInsets.fromLTRB(8.0, 0, 8.0, 0),
```
holtzmak, line 16: I think it would be nice to consider a CSS file with these constants, like `Colors, dimensions`, so we don't have [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) lying around:

- [x] Make a task to investigate and create a CSS/dimensions constants file. No need to fix right now.

_I think it is easier to understand using `EdgeInsets.fromLTRB(top_inset_for_widgets, etc...)` because it will save people time from playing around with sizes if we've already discovered the size. Additionally, it would be easier to remember such a CSS file over remembering that the top inset is 8.0 pixels._

app/lib/sidebar/local_widgets/list_tile.dart
```diff
@@ -0,0 +1,49 @@
...
+          onTap: () {
+            // TODO: Implement a better way of routing
+            Navigator.pop(context);
+            Navigator.push(
+              context,
+              MaterialPageRoute(
+                  builder: (context) => NavigationBarController()),
+            );
+          },
+          child: Row(
```
holtzmak, line 33: I wonder if these `children` are commonly used: do you think we could we extract these out in the future? 

I'm always thinking of refactoring... but my suggestions could be excessive - in the past I've certainly over-refactored things and made my code confusing for myself and others.

app/lib/sidebar/local_widgets/list_tile.dart
```diff
@@ -0,0 +1,49 @@
+import 'package:app/navbar/navigation.dart';
+import 'package:flutter/material.dart';
+
+class CustomTile extends StatelessWidget {
```
holtzmak, line 4: I wonder if we could have a better name than `CustomTile`. For example, your names `HomePage` `InProgressPage` `TravelHistory` are good names because I know what they represent right away. 

- [ ] What is a tile, and could we call this something like `QuickLinkTile`?

app/lib/screens/history/history.dart
```diff
@@ -1,6 +1,16 @@
 import 'package:flutter/material.dart';
 
-class TravelHistory extends StatelessWidget {
+class TravelHistory extends StatefulWidget {
+  @override
+  _TravelHistoryState createState() => _TravelHistoryState();
+}
+
+class _TravelHistoryState extends State<TravelHistory> {
```
holtzmak, line 8: Could you tell me what the differences between `state` `page` and `widget` are to you?

holtzmak (changes requested):
Some minor requests, questions from me. The drawer menu is a very useful thing we'll use regardless of our final GUI design.

I like the folder by feature organization; we should also be consistent with naming classes, widgets, etc. by feature too, and avoid generic names where we can. The reverse is also true, if we can make a reusable widget for several things, then we should give it a name like `TemplateWidget`.

app/lib/screens/home/home.dart
MummenRider: The build will run if state changes, but in this case it won't because it's stateless at the moment.  If you want state to persist, you need to use some kind of state management system. I included a todo inside main.dart for state management.

app/lib/sidebar/local_widgets/list_tile.dart
MummenRider: Yes, I saw git projects before where they separated all the styling. This wasn't the main concern of this commit and therefore did not bother doing it.

app/lib/sidebar/local_widgets/list_tile.dart
MummenRider: We can def extract widgets if necessary, but I don't think it was necessary for this specific commit.

app/lib/screens/home/home.dart
holtzmak: Then I'm thinking we could create the generic drawer before any other pages are made and used, and give this generic drawer to those pages which have the menu option. I think the drawer isn't meant to be stateful?

app/lib/sidebar/local_widgets/list_tile.dart
MummenRider: A tile is the links inside the drawer (e.g. Travel history). I'll looking into proper flutter naming convention in the future.  Since this my own implementation of ListTile widget, I decided to name it 'CustomTile'.  This is an example of extracting whatever is inside the child.

app/lib/screens/history/history.dart
MummenRider: Widget is the description of the UI (e.g. Padding., etc). Two categories of widgets, Stateless and Stateful. 
Stateless - UI/data that will never change.
Stateful - UI/data that can change.
I know in the future, the UI in travel history will change (e.g. adding travel).
Depending on your business logic, you need to decide whether a widget should be stateful or not. 

I encourage everyone (@holtzmak , @mansiPaTeLL , @SanaKhan1997 ) to look into different states of flutter and how to manage it.  Here's where to [start](https://flutter.dev/docs/development/data-and-backend/state-mgmt).

app/lib/screens/home/home.dart
MummenRider: Correct, Drawer widget should be stateless (as you can see in the code).

holtzmak: We're on the same page then.

- [x] When you go to make stateful classes, could you investigate injecting the drawer just once (meaning, every time the widget/or whatever is updated, that it doesn't need the drawer every time)

app/lib/sidebar/local_widgets/list_tile.dart
holtzmak: Oh, is this `child` a widget? I really need to read more...

app/lib/screens/home/home.dart
MummenRider: > We're on the same page then.
> 
> * [ ]  When you go to make stateful classes, could you investigate injecting the drawer just once (meaning, every time the widget/or whatever is updated, that it doesn't need the drawer every time)

I believe your concern can be resolve by using stack. I'm still new to flutter, but the stack I'm familiar is called IndexStack. I already implemented this in my bottom nav bar experimentation. The current state of the page will be persistent and won't need to re-build (no need to call BottomNavigation bar every time)
 I can explain this in further detail in our next meeting. Furthermore, this can also be solve by using some kind of [state management](https://flutter.dev/docs/development/data-and-backend/state-mgmt/options)

app/lib/sidebar/local_widgets/list_tile.dart
MummenRider: child is a named parameter which accept a type of widget <Widget>

holtzmak (approved):
Thanks for answering the questions I had. LGTM

Issue #100, opened 2021-01-08
Define the unit/integration/system test workflows 


holtzmak:
This task is done when we've defined a workflow/approach to writing code tests in Flutter that tells developers how to:
1. start writing tests
2. what they should consider when writing tests for the first time
3. and links to good TDD information to keep in mind when writing tests

This is especially important to help developers identify test cases "on the fly" for individual classes/widgets that aren't defined by system test procedure.

holtzmak:
@MummenRider , @SanaKhan1997 added the wiki page [Testing Workflow](https://github.com/holtzmak/Humane-Transport/wiki/Testing-Workflow). Let me know if you think something is missing.

MummenRider:
A really detailed wiki page, LGTM!

mansiPaTeLL:
@holtzmak I also reviewed it and it LGTM.

SanaKhan1997:
Looks good! thank you

PR #104, opened 2021-01-10
Add experimental sidebar integration test

### LINKED ISSUE #82: Experiment with Code Testing in Flutter
* Make a suite of tests that can run automatically (at the click of a button)
* Make and use a mock object in Dart (use a library like Mockito, but for Dart)
* Investigate if Flutter can do system level tests (i.e. once the simulator is running, interact with the GUI and verify events happen)
* Investigate if Flutter can do GUI tests (just the GUI code, like confirming that a button click starts an event)
* If the above two points are easy, write some simple test code

You should use an existing library(s): document what kind of options it provides use with (for example, can it use both android and iOS simulators? Can it mock external systems like databases?)

Any code you write to do the above can be merged into the codebase - we'll be able to overwrite/copy/etc. when we go to create the final GUI windows/widgets. You may want to keep a list of references that helped you: for example, Flutter cookbook pages - decide if we should have a dedicated wiki page for Flutter cookbook items.

holtzmak:
Closes #82 . 

Adds some configuration for Cocoapods and [Integration-Test plugin](https://flutter.dev/docs/testing/integration-tests). Note that this way of doing integration tests is a little different from the older Flutter driver-direct method in the Flutter Cookbook as of Dec 10, 2020.

Added a simple unit test to investigate what we'll need to do in order to grab things in the GUI for tests. We will absolutely need some stateful, key management @MummenRider in all our GUI classes with things we can interact with ([see this example as to why my test looks dumb and why we want to name widgets, we'll have a lot soon](https://stackoverflow.com/questions/53299088/how-to-open-a-drawer-in-flutter-test)).

Running this test manually is done like so, *inside the app directory*:
```
☁  app [experimental-integration-test] ⚡  flutter drive \
  --driver=test_driver/integration_test.dart \
  --target=integration_test/sidebar_test.dart  # Must list target to run!
# Not specifying a device with -d will use whatever's active, meaning my iPhone sim is open before I run the test
Using device iPhone 12 Pro Max.
Starting application: integration_test/sidebar_test.dart
Running Xcode build...                                                  
 └─Compiling, linking and signing...                        12.4s
Xcode build done.                                           35.5s
flutter: Observatory listening on http://127.0.0.1:58479/0KJsIuJ05a8=/
VMServiceFlutterDriver: Connecting to Flutter application at http://127.0.0.1:58479/0KJsIuJ05a8=/
VMServiceFlutterDriver: Isolate found with number: 1759583023509935
VMServiceFlutterDriver: Isolate is paused at start.
VMServiceFlutterDriver: Attempting to resume isolate
flutter: 00:00 +0: siderbar test tap on the siderbar; verify open sidebar
VMServiceFlutterDriver: Connected to Flutter application.
flutter: 00:01 +1: (tearDownAll)
# The following is a [known bug](https://github.com/flutter/flutter/issues/71594), ignore
flutter: Warning: integration_test test plugin was not detected.
flutter: 00:01 +2: All tests passed!
All tests passed.
Stopping application instance.
```

I will be writing a document as to how to start writing tests and how to run them in #100


MummenRider (approved):
LGTM!

PR #106, opened 2021-01-11
added database architecture folder

mansiPaTeLL:


holtzmak (changes requested):
- [x] Could you remove the "Second Copy of..."? I think we do not want this in the code base (if we do, please give it a better name!)

mansiPaTeLL:
@holtzmak Hey Kelly. Thanks for pointing that out. I have removed the file.

holtzmak (approved):

PR #108, opened 2021-01-12
Add user story maps

holtzmak:
While I was starting investigating system tests, I noticed we didn't have the USMs in `documentation`.

Here's two forms of what we have on [StoriesOnBoard](https://landofooo.storiesonboard.com/m/humane-transport) - you may have to zoom in a lot on the backlog format PDF but It's legible I promise.


mansiPaTeLL (approved):
Thanks @holtzmak LGTM.

Issue #99, opened 2021-01-08
Experiment/Investigate using a simple loading screen


holtzmak:
In contrast to #98 . If multithreading is too difficult for us to do in a reasonable amount of time, we can just make our application "single-threaded" (it isn't really, but we're not doing any additional threads). This means, when saving an ATR or some other potentially time-expensive action, we make the user wait until we've either successfully saved the ATR or failed to do so - we make them wait until we're done working.

This task is done when:
- [ ] We have a simple loading screen or pop-up that occurs when we're processing GUI actions, that doesn't lose the work for the previous screen (or underlying screen)

holtzmak:
Useful links:
1. [A whole screen dedicated to loading ](https://flutter.dev/docs/development/ui/advanced/splash-screen)
2. [Circular progress indicator or buffering icon](https://api.flutter.dev/flutter/material/CircularProgressIndicator-class.html)

holtzmak:
It appears that to make a blocking screen (preventing the user from interacting with the underlying widget), we can use the `barrierDismissible` property of `showDialog`. However, because the dialog is not dismissible, we'll need to hook up the `Navigator.pop()` to the process we're waiting on.

PR #109, opened 2021-01-13
Provider pattern

### LINKED ISSUE #105: Experiment on Provider Pattern for State Management
This task is done when:
- [x] Explore the needs of Provider pattern 
- [x] Provider pattern is integrated to our current application

*Note: This issue is continuation of issue #78

MummenRider:
[Source](https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple) for this commit .

By using provider pattern, I was able to use widget across the application. Furthermore, builds that are not necessary was also avoided. To check the persistency of each screen, I created a simple ListView widget that will represent user's travel history record. The travel history screen can still be refactored such as creating a reusable ListView widget and separate model class for its content. I did not do this because the main concern of this commit is to integrate provider pattern to our current application. Once we finalized our design, widgets that are reusable should go to `common/widgets` directory.

**Note to reviewers**: If you scroll down in travel history and stop at some point, then go to another screen and then go back to travel history page. The state of the list view will be the same (where the user stopped scrolling). To do this, I used IndexedStack. 
**Problem with this IndexStack approach**:
During the cold start of the application, widgets that are not visited yet will be builded. I believe this could be solve by looking into flutter route which is also has its own learning curve. For now, we only have tree items in bottom nav and therefore should not raise any concern.


MummenRider:
Fixing testing issues that is similar to [this](https://github.com/flutter/flutter/issues/15919).

app/lib/screens/history/history.dart
```diff
@@ -1,27 +1,29 @@
...
-      body: Center(
-        child: Text(
-          'This is Travel History page',
-          style: Theme.of(context).textTheme.headline5,
-        ),
+      body: ListView.builder(
+        itemCount: 20,
+        itemBuilder: (context, index) {
+          return Card(
+            color: index % 2 == 0 ? Colors.grey[300] : null,
```
holtzmak, line 12: Haha, alternating colours trick I see

app/test/main_test.dart
```diff
@@ -5,7 +5,7 @@ void main() {
   testWidgets('HumaneTransportApp says this is home page',
```
holtzmak: - [x] Please update the test name with changes

app/pubspec.lock
```diff
@@ -266,4 +280,4 @@ packages:
     version: "2.1.2"
 sdks:
   dart: ">=2.10.0-110 <2.11.0"
-  flutter: ">=1.12.13+hotfix.5 <2.0.0"
```
holtzmak, line 269: Did `flutter pub get` update this?

app/lib/screens/home/home.dart
```diff
@@ -1,37 +1,23 @@
...
-                  builder: (context) => NavigationBarController(),
-                ),
-              );
-            },
-            icon: Icon(Icons.add_circle_outline),
-            label: Text('New Travel'),
-          ),
-        ],
-      )),
+      bottomNavigationBar: BottomNavBar(),
```
holtzmak, line 20: Good refactoring example. I'm sure you're well aware of the following Clark - but to other readers:
1. We should extract all contextually different widgets/objects out to their own files
1. This adds to reusability and our understanding of composition
1. This supports the preference of composition over inheritance
1. [Flutter architecture is build with this in mind](https://flutter.dev/docs/resources/architectural-overview)

And, as a more personal note: this reduces file length and we like shallow files; not only does it make reading easier but because all contextually different code is separate, testing becomes easier.

app/lib/screens/home/home.dart
```diff
@@ -1,37 +1,23 @@
...
+import 'package:app/navigations/side_bar/sidebar.dart';
 import 'package:flutter/material.dart';
+import 'package:app/providers/navigation_provider.dart';
+import 'package:provider/provider.dart';
 
-class HomePage extends StatelessWidget {
+class HomeScreen extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
+    print('Application starting....');
```
holtzmak, line 10: Is it common to have these print statements in the final result? Maybe these are placeholders for like the buffering icon we intend in #99 ?

app/lib/screens/history/history.dart
MummenRider: I am pretty sure I learned this in cs 110 haha

holtzmak (changes requested):
Very nice to see these changes, and interesting how they're implemented. Following up with the [simple state management in the cookbook](https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple):
1. Interesting to see the argument against the `myObj.update()` and passing callback functions. These are very common methods of writing MVC patterns - so much so that one of my internships I wrote entirely on comparing the Observer pattern to these kinds of patterns (VERY similar to provider, but provider has a better name for it's context).
2. My results were that using `update()` was straightforward in the tell, don't ask paradigm and forces the class to figure out how to do the update. Particularly, controllers in the MVC pattern have this exact function to bubble up changes to GUI, and older GUI libraries use this pattern
3. However... Flutter appears to have underlying management that requires immutability, hence the `ChangeNotifier` which behaves like a Subject (Go4 pattern) or Observable (RX libraries). 

I wish we could add new records to tables and have them dynamically adjust without rebuilding the whole table - perhaps it's less intensive to swap with a new object than figure out how the GUI objects should adjust. I'm of the family that GUIs are not immutable but the data contained is (meaning the table is free game but the records in it aren't). Flutter takes this further.

ANYWAYS...

- [x] Does the comment in `main.dart` still apply for state management?

app/lib/screens/history/history.dart
holtzmak: I'd prefer to tell the table this in language than use a hacky trick. Is this the only way to do this as described by community?

app/test/main_test.dart
MummenRider: I forgot to remove that comment. I will remove the comment and update the test name in my next commit

app/pubspec.lock
MummenRider: Most likely, flutter pub get also updates dependencies

app/lib/screens/home/home.dart
MummenRider: Also, by extracting this widget it gives me the ability to determine if it's building or not

app/lib/screens/home/home.dart
MummenRider: These printing statements is for me to check if widget is build. In the future, we should use logs similar to .NET when I'm developing web apps. I'm pretty sure flutter has its own logging mechanism or even a plugin for it.

app/lib/screens/history/history.dart
MummenRider: Most likely, I just did this for fun and part of experimentation

app/lib/screens/home/home.dart
holtzmak: Yup, just noticed it in the build output now. We can add a task for logs later (when we come upon it)

holtzmak (approved):

PR #110, opened 2021-01-14
Pdf creation out of form element

### LINKED ISSUE #81: Experiment with making a PDF in Dart (Flutter)
* Make a PDF out of code objects using Dart
* Make a PDF out of code objects with images using Dart

You should use an existing library out there: document what kind of options it provides use with (for example, images, formatting the document, what about have a template PDF that's filled out using code objects for the ATR?)

Any code you write to do the above can be merged into the codebase - we'll be able to overwrite/copy/etc. when we go to create the final GUI windows/widgets. You may want to keep a list of references that helped you: for example, Flutter cookbook pages - decide if we should have a dedicated wiki page for Flutter cookbook items.

mansiPaTeLL:
Pdf-file-creation

### DIFF
--- a/app/ios/Podfile.lock
+++ b/app/ios/Podfile.lock
@@ -1,21 +1,33 @@
 PODS:
   - Flutter (1.0.0)
+  - flutter_full_pdf_viewer (1.0.1):
+    - Flutter
   - integration_test (0.0.1):
+    - Flutter
+  - path_provider (0.0.1):
     - Flutter
 
 DEPENDENCIES:
   - Flutter (from `Flutter`)
+  - flutter_full_pdf_viewer (from `.symlinks/plugins/flutter_full_pdf_viewer/ios`)
   - integration_test (from `.symlinks/plugins/integration_test/ios`)
+  - path_provider (from `.symlinks/plugins/path_provider/ios`)
 
 EXTERNAL SOURCES:
   Flutter:
     :path: Flutter
+  flutter_full_pdf_viewer:
+    :path: ".symlinks/plugins/flutter_full_pdf_viewer/ios"
   integration_test:
     :path: ".symlinks/plugins/integration_test/ios"
+  path_provider:
+    :path: ".symlinks/plugins/path_provider/ios"
 
 SPEC CHECKSUMS:
   Flutter: 0e3d915762c693b495b44d77113d4970485de6ec
+  flutter_full_pdf_viewer: 1b65139fc3236cde300ba6bea2a72669cb07f2e6
   integration_test: 5ed24a436eb7ec17b6a13046e9bf7ca4a404e59e
+  path_provider: abfe2b5c733d04e238b0d8691db0cfd63a27a93c
 
 PODFILE CHECKSUM: aafe91acc616949ddb318b77800a7f51bffa2a4c
 
--- a/app/lib/screens/history/history.dart
+++ b/app/lib/screens/history/history.dart
@@ -1,4 +1,8 @@
+import 'dart:io';
+
+import 'package:app/screens/history/pdf_preview.dart';
 import 'package:flutter/material.dart';
+import 'package:path_provider/path_provider.dart';
 
 class TravelHistory extends StatelessWidget {
   @override
@@ -24,6 +28,26 @@ class TravelHistory extends StatelessWidget {
             ),
           );
         },
+      ),
+      floatingActionButton: FloatingActionButton(
+        onPressed: () async {
+          try {
+            buildPdf();
+            savePdf();
+            final output = await getTemporaryDirectory();
+            String fullPath = ("${output.path}/example.pdf");
+
+            Navigator.push(
+                context,
+                MaterialPageRoute(
+                    builder: (context) => PdfPreview(
+                          path: fullPath,
+                        )));
+          } on HttpException catch (err) {
+            print('Error: $err');
+          }
+        },
+        child: Icon(Icons.save),
       ),
     );
   }
--- a/app/pubspec.yaml
+++ b/app/pubspec.yaml
@@ -23,8 +23,13 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+
   cupertino_icons: ^1.0.0
   provider: ^4.3.2+3
+
+  pdf: ^2.0.0
+  flutter_full_pdf_viewer: ^1.0.6
+  path_provider: ^1.5.1
 
 dev_dependencies:
   flutter_driver:
@@ -32,6 +37,7 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   integration_test: ^1.0.0
+
 
 # For information on the generic Dart part of this file, see the
 # following page: https://dart.dev/tools/pub/pubspec
--- a/app/lib/screens/history/pdf_preview.dart
+++ b/app/lib/screens/history/pdf_preview.dart
@@ -1,0 +1,59 @@
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_full_pdf_viewer/full_pdf_viewer_scaffold.dart';
+import 'package:path_provider/path_provider.dart';
+import 'package:pdf/pdf.dart';
+import 'package:pdf/widgets.dart' as pw;
+
+class PdfPreview extends StatelessWidget {
+  final String path;
+
+  PdfPreview({this.path});
+
+  @override
+  Widget build(BuildContext context) {
+    return PDFViewerScaffold(
+      appBar: AppBar(
+        title: Text('Back to Home'),
+      ),
+      path: path,
+    );
+  }
+}
+
+//Create a blank document
+final pdf = pw.Document();
+//add data into the pdf file
+buildPdf() {
+  pdf.addPage(pw.MultiPage(
+    pageFormat: PdfPageFormat.a4,
+    margin: pw.EdgeInsets.all(34),
+    build: (pw.Context context) {
+      return <pw.Widget>[
+        /*
+              We will be calling API and fetching data from the database here.
+              Now just adding random text for experimentation purpose. Once we have database/API set up, will
+              experimenting on fetching data
+              */
+        pw.Header(level: 0, child: pw.Text('Shipper Information')),
+        pw.Paragraph(text: 'Some Information'),
+        pw.Paragraph(text: 'Some Information'),
+        pw.Header(level: 1, child: pw.Text("Travel Information")),
+        pw.Paragraph(text: 'Some Text 1'),
+        pw.Paragraph(text: 'Some Text 2')
+      ];
+    },
+  ));
+}
+
+Future savePdf() async {
+  try {
+    final output = await getTemporaryDirectory();
+    final file = File("${output.path}/example.pdf");
+    file.writeAsBytesSync(await pdf.save());
+  } on HttpException catch (err) {
+    print('Error: $err');
+  }
+}
--- a/app/pubspec.lock
+++ b/app/pubspec.lock
@@ -22,6 +22,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.5.0-nullsafety.1"
+  barcode:
+    dependency: transitive
+    description:
+      name: barcode
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "1.17.1"
   boolean_selector:
     dependency: transitive
     description:
@@ -85,6 +92,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "1.2.0-nullsafety.1"
+  ffi:
+    dependency: transitive
+    description:
+      name: ffi
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "0.1.3"
   file:
     dependency: transitive
     description:
@@ -102,6 +116,13 @@ packages:
     description: flutter
     source: sdk
     version: "0.0.0"
+  flutter_full_pdf_viewer:
+    dependency: "direct main"
+    description:
+      name: flutter_full_pdf_viewer
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "1.0.6"
   flutter_test:
     dependency: "direct dev"
     description: flutter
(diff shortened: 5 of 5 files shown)


app/lib/screens/history/history.dart
```diff
@@ -23,6 +60,22 @@ class _TravelHistoryState extends State<TravelHistory> {
...
+
+          Navigator.push(
+              context,
+              MaterialPageRoute(
+                  builder: (context) => PdfPreview(
+                        path: fullPath,
+                      )));
+        },
+        child: Icon(Icons.save),
+      ),
```
MummenRider: Is travel history the only screen where we can generate pdfs? If so, we can extract this widget and put it in `lib/screens/local_wdget`

app/lib/screens/history/history.dart
```diff
@@ -11,6 +18,36 @@ class _TravelHistoryState extends State<TravelHistory> {
...
+          pw.Paragraph(text: 'Some Text 2'),
+        ];
+      },
+    ));
+  }
+
+  Future savePdf() async {
+    final output = await getTemporaryDirectory();
+    final file = File("${output.path}/example.pdf");
+    file.writeAsBytesSync(await pdf.save());
```
MummenRider: I think this applies to any async/await calls, it's best to wrap this with try/catch block to handle exception errors

app/ios/Podfile.lock
```diff
@@ -1,21 +1,33 @@
...
+  - flutter_full_pdf_viewer (1.0.1):
+    - Flutter
   - integration_test (0.0.1):
     - Flutter
+  - path_provider (0.0.1):
+    - Flutter
 
 DEPENDENCIES:
   - Flutter (from `Flutter`)
+  - flutter_full_pdf_viewer (from `.symlinks/plugins/flutter_full_pdf_viewer/ios`)
```
MummenRider, line 12: I have concern about this plugin:  flutter_full_pdf_viewer. To view the pdf, the plugin created its own Scaffold but I believe it's missing important components to modify the UI. I checked the plugin and it seems there's [issues](https://pub.dev/packages/flutter_full_pdf_viewer/score) in it. 

- [ ] Have you tried other pdf viewer plug in?

app/lib/screens/history/PdfPreview.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_full_pdf_viewer/flutter_full_pdf_viewer.dart';
+
+class PdfPreview extends StatelessWidget {
```
holtzmak: - [x] The filename should be lowercase `pdfpreview.dart` or `pdf_preview.dart`

app/lib/screens/history/history.dart
```diff
@@ -11,6 +18,36 @@ class _TravelHistoryState extends State<TravelHistory> {
     print('Init: Travel history');
   }
 
+  final pdf = pw.Document(); //Create a blank document
```
holtzmak: As per our previous discussion we should extract this into its own file, but I think it would go well in `pdf_preview.dart`.

- [x] Move out of `history.dart`

app/lib/screens/history/history.dart
holtzmak: Not sure if related to this comment or [this one](https://github.com/holtzmak/Humane-Transport/pull/110/files?file-filters%5B%5D=.dart&file-filters%5B%5D=.lock#r556983781), but I had the simulator open with your PDF page open, and my machine started to sound like it was about to launch into outer space.

- [x] Check that generating this PDF doesn't also do this to your machine. Running the application with Clark's recent PR did not jump my machine into hyperspeed

app/lib/screens/history/history.dart
```diff
@@ -11,6 +18,36 @@ class _TravelHistoryState extends State<TravelHistory> {
...
+          pw.Paragraph(text: 'Some Information'),
+          pw.Header(level: 1, child: pw.Text("Travel Information")),
+          pw.Paragraph(text: 'Some Text 1'),
+          pw.Paragraph(text: 'Some Text 2'),
+        ];
+      },
+    ));
+  }
+
+  Future savePdf() async {
```
holtzmak: I see this warning in my IDE upon inspecting your code:
Screen Shot 2021-01-13 at 7 59 59 PM

- [x] Please fix or find out why

app/lib/screens/history/history.dart
```diff
@@ -23,6 +60,22 @@ class _TravelHistoryState extends State<TravelHistory> {
           style: Theme.of(context).textTheme.headline5,
         ),
       ),
+      floatingActionButton: FloatingActionButton(
+        onPressed: () async {
+          buildPdf();
+          savePdf();
+          final output = await getTemporaryDirectory();
```
holtzmak: I see this warning in my IDE upon inspecting your code:
Screen Shot 2021-01-13 at 8 00 11 PM

- [x] Please fix or find out why

app/lib/screens/history/history.dart
```diff
@@ -11,6 +18,36 @@ class _TravelHistoryState extends State<TravelHistory> {
     print('Init: Travel history');
   }
 
+  final pdf = pw.Document(); //Create a blank document
+  //add data into the pdf file
+  buildPdf() {
```
holtzmak: I happen to see this in the build output now:
```
flutter: Helvetica has no Unicode support see https://github.com/DavBfr/dart_pdf/wiki/Fonts-Management
```

- [ ] Please fix or find out why

app/lib/screens/history/history.dart
holtzmak: Probably not only in `TravelHistory`. There could be a case where a partial form could be emailed. Maybe this can go under `common/utility`

holtzmak (changes requested):
In conjunction with Clark's questions, please address!

app/lib/screens/history/history.dart
mansiPaTeLL: No. But I can create different class to generate the pdf file. I did this just for testing purposes. This class will go in different folder and we can use that class whenever we need it. I put it in here because I was thinking that whenever traveller wants to access past `TravelHistory` & when click on the particular document, they can view in pdf format and on the same page we can have share option using email (or any other way).

app/lib/screens/history/history.dart
mansiPaTeLL: > I think this applies to any async/await calls, it's best to wrap this with try/catch block to handle exception errors

Yes, during my research I also found that if you want to have exception handling in async function we can defiantly do that. I have added exception handling with async/await.

app/ios/Podfile.lock
mansiPaTeLL: @MummenRider So, I haven't used any other plugins, as this is most suggested library to use while I was researching. If you look at the scores, you will see that that it doesn't support the web platform but it supports Android/iOS platform perfectly fine. For the statistical analysis, it says it has minor formatting file issue. But, they have provided the solution for it as well. It has 96% popularity as well. I think this will be fine. I also found other library like pdf viewer plugin but it has less popularity than this one. S oI preferred to use flutter_full_pdf_viewer. 

Also, I am not sure why it is running `- flutter_full_pdf_viewer (from `.symlinks/plugins/flutter_full_pdf_viewer/ios`)` in Podfile.lock file. Ans similar thing is happening with integration_test & path_provider dependencies as well. [Check it out her](https://github.com/holtzmak/Humane-Transport/pull/110/files?file-filters%5B%5D=.dart&file-filters%5B%5D=.lock#r556983781)

app/lib/screens/history/PdfPreview.dart
mansiPaTeLL: @holtzmak I changed the file name. Thank you.

app/lib/screens/history/history.dart
mansiPaTeLL: @holtzmak I moved this into pdfpreview.dart file. Thanks.

app/lib/screens/history/history.dart
mansiPaTeLL: @holtzmak I don't find this warning into my IDE. Where exactly you get this warning? I checked in my problem window I have no error or warnings. I think you need to go to `pubspec.yaml` file and do `pub get` & `pub upgrade` and restart your IDE.  I think this should resolve the issue.  
I am assuming that your IDE not detecting the package.

app/lib/screens/history/history.dart
mansiPaTeLL: @holtzmak I don't see this into my IDE again. I think once go to `pubspec.yaml` file and do `pub get` & `pub upgrade` snd restart IDE, will be able to resolve the issue.

app/lib/screens/history/history.dart
holtzmak: I did not try `flutter pub upgrade`: either my IDE was out of date or this fixed it. No problem here.

app/ios/Podfile.lock
MummenRider: flutter_full_pdf_viewer plugin is actually not bad. What I don't like is they limit the methods we can use. At the moment, PDFViewerScaffold only have the following:
1.) App bar customization
2.) The path
3.) A type bool called primary (used in appbar customization)

Problem with this is it do not accept child where we can add more widgets for body customization. As of now, user needs to scroll down to check other pages by using this plug in. How can we add UI where the user can simply click 'next' button to see other pages instead of scrolling down?

mansiPaTeLL: @mansiPaTeLL I got your concern. There are so many pdf viewer plugging available to use. And it easy to use clark. for example, `https://pub.dev/packages/advance_pdf_viewer.` We are not sticking with particular plugin. If you want to use any other plugins you can free to use. I am just doing experimenting right now and we are not sticking with particular option. I hope this helps you. Thanks. 

If you want to decorate the pdf file it self, there are so many functions available  with pdf widget plugins. Thanks.

MummenRider: I see, it's good to know that we're not stuck to this specific plugin

app/lib/screens/history/history.dart
MummenRider: @holtzmak @mansiPaTeLL I believe this is because the font Helvetica is not included in pubspec.yaml

app/ios/Podfile.lock
mansiPaTeLL: @MummenRider Yes, but if you also look at the score for other plugins, they have less popularity so before using other plugins, we need to make sure it is compatible. But always if you find the better one, feel free to use it.

app/lib/screens/history/history.dart
mansiPaTeLL: @holtzmak @MummenRider I tried to fix this by using suggestion guide. But then it is giving me some exception handling error. I think this can be considered as a separate task as it takes some more research.

MummenRider: @mansiPaTeLL , try this [solution](https://github.com/DavBfr/dart_pdf/issues/196)

mansiPaTeLL: @MummenRider Thanks for this solution. I already looked into this solution. I tried and I am having an issue with loading `assets/open-sans.ttf`

MummenRider (commented):
This commit's focus is to explore on how to implement PDF creation where I believe @mansiPaTeLL accomplished. However, one thing I want to point out is the separation of business logic to app logic. Right now, this commit exposed the business logic (e.g. buildPdf and savePdf) to the UI component. Do note that we will put all business logics inside the provider folder I recently merged to master. To start, I suggest you to read [this](https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple)

mansiPaTeLL:
> This commit's focus is to explore on how to implement PDF creation where I believe @mansiPaTeLL accomplished. However, one thing I want to point out is the separation of business logic to app logic. Right now, this commit exposed the business logic (e.g. buildPdf and savePdf) to the UI component. Do note that we will put all business logics inside the provider folder I recently merged to master. To start, I suggest you to read [this](https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple)

Thank you for your suggestion. I will defiantly read it out. I also believe that we need to defiantly separate the business logic and app logic. I saw that you have just merged provider folder to add business logics code. For my next task, I will add all business related code to provider folder. Thank you.

app/lib/screens/history/PdfPreview.dart
holtzmak: I do not see a change in the file name?

app/lib/screens/history/history.dart
```diff
@@ -23,6 +27,26 @@ class _TravelHistoryState extends State<TravelHistory> {
           style: Theme.of(context).textTheme.headline5,
         ),
       ),
+      floatingActionButton: FloatingActionButton(
+        onPressed: () async {
+          try {
+            buildPdf();
```
holtzmak, line 35: And yes, as per Clark's prior comment please be aware this is to be revisited.

app/pubspec.yaml
```diff
@@ -25,17 +25,26 @@ dependencies:
     sdk: flutter
 
 
+
```
holtzmak: These extra lines look a little odd

holtzmak (changes requested):
Just wondering about the file name change. 

Additionally, you marked off a previous comment of mine about leaving the app with PDF open causing more RAM to be consumed. Did you happen to see this on your machine as well? What about you @MummenRider ?

MummenRider:
> Just wondering about the file name change.
> 
> Additionally, you marked off a previous comment of mine about leaving the app with PDF open causing more RAM to be consumed. Did you happen to see this on your machine as well? What about you @MummenRider ?

I did not experience this at all.

app/lib/screens/history/PdfPreview.dart
mansiPaTeLL: I did change the file name, I don't know why it is not showing. I tried to push again but it is giving me message that everything is up-to-date. 
image

app/lib/screens/history/history.dart
mansiPaTeLL: > Just wondering about the file name change.
> 
> Additionally, you marked off a previous comment of mine about leaving the app with PDF open causing more RAM to be consumed. Did you happen to see this on your machine as well? What about you @MummenRider ?

I did not experienced any even I have 4GB RAM.

holtzmak: OK, I imagine this is my personal problem then.

holtzmak (approved):

PR #111, opened 2021-01-14
Adding email sender

### LINKED ISSUE #79: Experiment with Sending Emails with Flutter
* Make the application send a (no-reply) email

Any code you write to do the above can be merged into the codebase - we'll be able to overwrite/copy/etc. when we go to create the final email interface API. You may want to keep a list of references that helped you: for example, Flutter cookbook pages - decide if we should have a dedicated wiki page for Flutter cookbook items.

SanaKhan1997:
For now I created a form that will allow to take recipient email to take to gmail.

### DIFF
--- a/app/lib/email_sender.dart
+++ b/app/lib/email_sender.dart
@@ -1,7 +1,6 @@
 import 'package:flutter_email_sender/flutter_email_sender.dart';
 import 'package:flutter/material.dart';
 
-
 class EmailSender {
   Future<void> send(
       String emailBody, String eSubject, String eRecipients) async {
@@ -13,7 +12,6 @@ class EmailSender {
       isHTML: false,
     );
     String emailResponse;
-
 
     try {
       await FlutterEmailSender.send(email);
@@ -31,8 +29,6 @@ class EmailRoute extends StatelessWidget {
 
   final validate = false;
 
-
-
   final email = EmailSender();
   final recipientController = TextEditingController();
   final subjectController = TextEditingController();
@@ -49,11 +45,9 @@ class EmailRoute extends StatelessWidget {
               key: formKey,
               child: Column(children: <Widget>[
                 TextFormField(
-
                   decoration: InputDecoration(
                       labelText: 'Enter Email',
                       errorText: validate ? 'Empty email address' : null),
-
                   controller: recipientController,
                 ),
                 TextFormField(
--- a/app/lib/email_sender.dart
+++ b/app/lib/email_sender.dart
@@ -1,0 +1,77 @@
+
+import 'package:flutter_email_sender/flutter_email_sender.dart';
+import 'package:flutter/material.dart';
+
+
+class EmailSender {
+  Future<void> send(
+      String emailBody, String eSubject, String eRecipients) async {
+    final Email email = Email(
+      body: emailBody,
+      subject: eSubject,
+      recipients: [eRecipients],
+      attachmentPaths: null,
+      isHTML: false,
+    );
+    String emailResponse;
+    try {
+      await FlutterEmailSender.send(email);
+      emailResponse = 'success';
+    } catch (error) {
+      emailResponse = error.toString();
+    }
+
+    print(emailResponse);
+  }
+}
+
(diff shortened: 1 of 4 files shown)


app/lib/email_sender.dart
```diff
@@ -0,0 +1,74 @@
...
+              key: formKey,
+              child: Column(children: <Widget>[
+                TextFormField(
+                  decoration: InputDecoration(hintText: 'Email'),
+                  controller: recipientController,
+                  validator: (value){
+                    if(value.isEmpty){
+                      return 'Please enter email';
+                    }else
+                      return null;
```
holtzmak: We want to avoid using null in Dart (as with most languages, Java had originally introduced it's abuse and it causes bugs all the time...) see [here for options on how we could refactor this to be non-null or at least safe](https://dart.dev/null-safety).

- [x] Make this null-safe if possible

app/lib/email_sender.dart
```diff
@@ -0,0 +1,74 @@
...
+            child: Form(
+              key: formKey,
+              child: Column(children: <Widget>[
+                TextFormField(
+                  decoration: InputDecoration(hintText: 'Email'),
+                  controller: recipientController,
+                  validator: (value){
+                    if(value.isEmpty){
+                      return 'Please enter email';
+                    }else
```
holtzmak: Seeing that some space is missing here, you'll need to run `flutter format .` on your code and make another commit with the changes (this is probably why your checks are failing).
Sorry, pre-commits aren't set up for windows yet.

- [x] Run `flutter format .`

app/lib/email_sender.dart
```diff
@@ -0,0 +1,74 @@
...
+                    }
+                  },
+                  child: Text('Submit'),
+                ),
+              ]),
+            )
+        )
+    );
+  }
+}
```
holtzmak: - [x] Missing EOL, maybe `flutter format .` will fix

holtzmak (changes requested):
Minor requested changes and likely where GitHub actions is failing. Sorry about `pre-commit` and windows... you'll either have to manually remember these things or mess around with git hooks manually. Up to you.

app/lib/email_sender.dart
MummenRider: I know that this commit is to explore email sending feature, but do note that there should be only one class in one file. In the future (no need to change), EmailSender can go to model directory and email route can go to navigations directory.

app/lib/email_sender.dart
```diff
@@ -0,0 +1,73 @@
...
+                ),
+                RaisedButton(
+                  onPressed: () {
+                    if (formKey.currentState.validate()) {
+                      print('valid');
+                      email.send(bodyController.text, subjectController.text,
+                          recipientController.text);
+                    }
+                  },
+                  child: Text('Submit'),
```
MummenRider, line 69: Once we started real development (no need to change rn), as for best practice, we should create custom widgets so we can use it anywhere on the app. We should make things reusable.

app/lib/email_sender.dart
```diff
@@ -0,0 +1,73 @@
...
+                  decoration: InputDecoration(hintText: 'Message'),
+                  controller: bodyController,
+                ),
+                RaisedButton(
+                  onPressed: () {
+                    if (formKey.currentState.validate()) {
+                      print('valid');
+                      email.send(bodyController.text, subjectController.text,
+                          recipientController.text);
+                    }
```
MummenRider, line 67: How do we handle formkey if it's not valid?

app/lib/email_sender.dart
SanaKhan1997: changed form validation to null safe expression

app/lib/email_sender.dart
SanaKhan1997: Alright!

app/lib/email_sender.dart
SanaKhan1997: If not valid form key will not proceed with submission

MummenRider (approved):
LGTM!

holtzmak (approved):

Issue #113, opened 2021-01-16
Keep Bottom Nav Persistent across the application


MummenRider:
As shown on the video, bottom nav disappear when new screen is pushed. 

![bottomNavBug](https://user-images.githubusercontent.com/43891568/104816258-925faf80-57df-11eb-8995-707238724c18.gif)

mansiPaTeLL:
@MummenRider I think this is required to solve. Thanks for creating this issue.

holtzmak:
EDIT: Made my description more concise yet again

This bug is caused by the following:
1. that we set up our `NavigationProvider` with the expected three screens
1. but we tell the consumer in `home.dart` that those screens which consume (use) the `NavigationProvider` are only those screens in the `NavProvider`:

I'll experiment with a solution like [this one](https://medium.com/@bolajifemi28/flutter-how-to-make-your-bottom-nav-bar-visible-persistent-when-navigating-to-a-new-route-3fac3e935cd7) on our `PdfPreview` widget (because it doesn't have a nav bar at the moment).

@MummenRider can you tell me, with your experience, if the above solution contradicts the provider pattern widget tree?

MummenRider:
@holtzmak the bug exists because at the moment, we don't have proper navigation routing where everything is an entry point. 
Right now, our routing look like '/new_travel' and when the button is pushed, we're pushing another page and the routing will look like '/test_screen'. To fix this issue, I think, the routing should look like '/new_travel/test_screen' when the button is pushed. 

The link you provided uses a 'ready made' bottom navigation bar called custom_navigator. Personally,  I try to avoid adding plugins when I can do it myself. I'm pretty sure we can do this without depending on any plugin. We just need to study a little bit more.

And yes, the code in that article do not use Provider pattern.

holtzmak:
After a lengthy side discussion with @MummenRider , we've come tentatively to the decision that having a persistent navigation bar with a deep widget tree is a bad idea. We'll be discussing later how we (the team) should handle navigation with the following reasons in mind:
 1. The `Navigator` affects how OS navigation is done (hitting the back button will cycle through the `IndexedStack`)
 2.  Maintaining a list of all possible routes in our application introduces exponential boiler plating: [this example](https://stackoverflow.com/questions/49681415/flutter-persistent-navigation-bar-with-named-routes) also brings up the point that a nav for every window is anti-pattern/bad practice.
[...]

PR #114, opened 2021-01-17
Refactor pdf preview for simple navigation

holtzmak:
Related to #113 .

I've refactored the PDF preview class to better use asynchronous code. This was also to play with `Future`s in dart.

I think following the discussion of #113 that any non-nav widget should use `Navigator.push(context, widget(any updated fields)` to push new screens without the nav bar attached, as seen in my changes to `history.dart`.

EDIT: Here's what the changes look like (not much)

https://user-images.githubusercontent.com/32527219/104831071-df2ca000-584a-11eb-9641-89f0c13a4fb6.mov


app/lib/common/utilities/pdf_preview.dart
```diff
@@ -0,0 +1,73 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_full_pdf_viewer/full_pdf_viewer_scaffold.dart';
+import 'package:path_provider/path_provider.dart';
+import 'package:pdf/pdf.dart';
+import 'package:pdf/widgets.dart' as pw;
+
+class PdfPreview extends StatefulWidget {
+  @override
```
holtzmak, line 11: We'll be able to pass in the ATR object in the future like so:
```suggestion
  final TransportRecord record;
  
  @override
```
And then access the record in `initState()` to update itself.

app/lib/common/utilities/pdf_preview.dart
```diff
@@ -0,0 +1,73 @@
...
+    _pdf = _getBuiltAndSavedPdf();
+  }
+
+  @override
+  build(BuildContext context) {
+    return FutureBuilder<File>(
+        future: _pdf,
+        builder: (BuildContext context, AsyncSnapshot<File> snapshot) {
+          switch (snapshot.connectionState) {
+            case ConnectionState.waiting:
```
holtzmak, line 30: We need this because the `Future` might not be ready to check if it's successful or failed yet.

app/lib/common/utilities/pdf_preview.dart
```diff
@@ -0,0 +1,73 @@
...
+          pw.Paragraph(text: 'Some Text 1'),
+          pw.Paragraph(text: 'Some Text 2')
+        ];
+      },
+    ));
+
+    var temporaryDirectory = await getTemporaryDirectory();
+    // TODO: Change the filename to something useful
+    return File("${temporaryDirectory.path}/example.pdf")
+        .writeAsBytes(await pdf.save());
```
holtzmak, line 71: `writeAsBytes` also returns a future, so if any of `pdf.save()`, `writeAsBytes`, `File` were to fail, that error will bubble up to the `FutureBuilder` to show.

app/lib/providers/navigation_provider.dart
```diff
@@ -10,7 +10,9 @@ class NavigationProvider extends ChangeNotifier {
     OngoingTravel(),
     TravelHistory(),
   ];
+
```
holtzmak, line 13: Interesting that `flutter format` didn't do this by my IDE, with the same settings, did

app/lib/screens/history/history.dart
```diff
@@ -30,23 +27,8 @@ class TravelHistory extends StatelessWidget {
...
-                MaterialPageRoute(
-                    builder: (context) => PdfPreview(
-                          path: fullPath,
-                        )));
-          } on HttpException catch (err) {
-            print('Error: $err');
-          }
-        },
+        onPressed: () => Navigator.push(
+            context, MaterialPageRoute(builder: (context) => PdfPreview())),
```
holtzmak, line 31: We can give the ATR object like so when the time comes
```suggestion
            context, MaterialPageRoute(builder: (context) => PdfPreview(record = selectedRecord))),
```

app/lib/common/utilities/pdf_preview.dart
```diff
@@ -0,0 +1,73 @@
...
+                    title: Text('Example PDF'),
+                  ),
+                  path: snapshot.data.path,
+                );
+          }
+        });
+  }
+
+  // TODO: Give this function an object to build PDF from
+  Future<File> _getBuiltAndSavedPdf() async {
```
holtzmak, line 48: I think normally these would be separate functions, but I don't see why we'd ever build a PDF to then not save/use it.

app/lib/screens/history/history.dart
MummenRider, line 31: In the future, we can further refactor  the navigation by using onGenerateroute feature in flutter. We can even pass data as named parameters which is even more cleaner. Thanks kelly!

MummenRider (approved):
All changes made here LGTM!

app/lib/screens/history/history.dart
holtzmak: I believe the named parameters require a mapping of name to widget somewhere, is that right?

PR #115, opened 2021-01-19
Improved navigation

### LINKED ISSUE #113: Keep Bottom Nav Persistent across the application

MummenRider:
- This commit includes nested navigation. Each tab will serve as parent nav and everything inside will be the inner nav. 
- For now, I moved email_sender inside utilities. 
- The counter I implemented in one of the screen did not use provider pattern. It's there to represent data persistency despite of going to other screens
- Sources:

1. [Flutter Documentation](https://api.flutter.dev/flutter/widgets/Navigator-class.html#nesting-navigators) 
2. [Stack Overflow](https://stackoverflow.com/questions/49681415/flutter-persistent-navigation-bar-with-named-routes)

- Video below shows the current navigation

https://user-images.githubusercontent.com/43891568/104996223-19ce3e00-59ed-11eb-9090-b5d13f3a1e60.mov


app/lib/routes/main_navigator/app_route.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:app/screens/home/home.dart';
+import 'package:app/screens/new_travel/new_travel_journey/test_screen_one.dart';
+import 'package:app/screens/new_travel/new_travel_journey/test_screen_two.dart';
+
+import 'package:flutter/material.dart';
+
+class AppRouteGenerator {
+  static Route<dynamic> onGenerateRoute(RouteSettings settings) {
```
MummenRider, line 8: AppRouteGenerator contains our top level navigator. Though the 2 test screens are part of the inner navigation, it's still needed here to give us an option whether bottom nav will persist or not.

app/lib/screens/history/history.dart
```diff
@@ -27,8 +30,7 @@ class TravelHistory extends StatelessWidget {
         },
       ),
       floatingActionButton: FloatingActionButton(
-        onPressed: () => Navigator.push(
-            context, MaterialPageRoute(builder: (context) => PdfPreview())),
+        onPressed: () => Navigator.of(context).pushNamed(PdfPreview.route),
```
MummenRider, line 33: Cleaner version when navigating through pages.

app/ios/Podfile.lock
```diff
@@ -1,5 +1,7 @@
 PODS:
   - Flutter (1.0.0)
+  - flutter_email_sender (0.0.1):
```
holtzmak, line 3: Ah yes, I suppose we missed testing this on iOS sim on local and checking that in. I suppose the GitHub Actions machine did this, but of course, that's lost to us.

holtzmak: I think it is fine to just have the next person commit these changes vs. Sana asking someone to see if cocoa pods changed from her non-mac changes.

app/lib/models/nav_items.dart
```diff
@@ -1,14 +1,26 @@
 import 'package:flutter/material.dart';
 
-class NavigationItems {
+/// [NavigationItem] holds all information when building [BottomNavigationBarItem].
```
holtzmak, line 3: Not sure I would keep these comments, I think it obvious from the code what each is doing?

app/lib/providers/navigation_provider.dart
```diff
@@ -1,22 +1,47 @@
...
+    NavigationItem(
+      label: 'Travel History',
+      icon: Icons.timeline_outlined,
+      initialRoute: TravelHistory.route,
+      onGenerateRoute: HistoryRouteGenerator.onGenerateRoute,
+      navigatorState: GlobalKey<NavigatorState>(),
+    )
   ];
 
+  // Getters
```
holtzmak, line 37: Definitely don't need these comments, were they generated?

app/lib/routes/main_navigator/app_route.dart
```diff
@@ -0,0 +1,29 @@
...
+            return HomeRootScreen();
+
+          case TestScreenOne.route:
+            return TestScreenOne();
+
+          case TestScreenTwo.route:
+            return TestScreenTwo();
+
+          default:
+            throw Exception('Invalid route: ${settings.name}');
```
holtzmak, line 24: - [x] Do exceptions crash the application or provide a pop-up at all (notify the user without looking at build output)?

app/lib/screens/new_travel/new_travel_journey/test_screen_one.dart
```diff
@@ -0,0 +1,54 @@
...
+              onPressed: () => Navigator.of(context).pop(),
+              child: Text('Go Back'),
+            ),
+            SizedBox(
+              height: 80.0,
+            ),
+            Text('Counter is $_counter'),
+            RaisedButton(
+              onPressed: () {
+                setState(() {
```
holtzmak, line 43: Curious, what happens if we don't go through `setState()`, does the widget not rebuild?

holtzmak (approved):
Changes LGTM! Only possible change would be from the behaviour of throwing exceptions - if the exception just crashes the app (which they do in Java) that's no good, I would personally prefer to let the app eventually degrade (if that's how bad routes are handled) and notify the user vs. just close the app.

app/lib/routes/main_navigator/app_route.dart
MummenRider: This is actually helpful during development. In production, we will have dedicated screen that will handle undefined routes. For now, the system will crash but it gives you useful info

app/lib/screens/new_travel/new_travel_journey/test_screen_one.dart
MummenRider: Correct! As we all know, Flutter is declarative which means UI depends on the data/function we pass. In this case, we need to change 0 to 1 and for that to work, we need to rebuild the widget. That is how flutter works. If you want something to change in UI, then you need to rebuild the widget

app/lib/models/nav_items.dart
MummenRider: Will remove unnecessary comments. I was just following the documentation and they do it like this ahaha

Issue #80, opened 2020-12-18
Experiment with Flutter Database VPN Connections


holtzmak:
Using the project framework set up in #56 , and useful database connection API in #70 , experiment with the following items:
* Make a VPN connection to our database, then store information and retrieve information over the connection
* May need to set our database to use VPN if not done already

Any code you write to do the above can be merged into the codebase - we'll be able to overwrite/copy/etc. when we go to create the final database interface API. You may want to keep a list of references that helped you: for example, Flutter cookbook pages - decide if we should have a dedicated wiki page for Flutter cookbook items.

holtzmak:
- [x] Follow up with project mentor about encryption schemes for connection and data storage

holtzmak:
Response from project mentor, scrubbed:

"I don't think it's necessary to encrypt the database, as the TSL encryption provided by https is sufficient for a large category of use cases.

Yes it is possible to circumvent https protections, but it requires considerable expertise, resources, and perseverance.  Https is a very well-designed and successful protocol, even nation-state-level actors find it a formidable obstacle. 

Another consideration is effort vs. payoff.  Adding another layer of security, such as a second layer of encryption, should be justified in terms of a threat model.  What are you trying to protect?  Who are the adversaries, and what harms could result from their various activities? For a majority of systems, https is sufficient."

- [x] We should follow-up with the client (or review past communications) for exactly how private ATR information should be.

_I have personal reason to believe that the most private information we have is transporter account information, but if we apply access-based rules on the database for what accounts read/write to the database, we can design access to the database to be as robust as it needs to. We can't stop a transporter from handing out their personal login information, but we can limit the harm from such an action on others using the system._

holtzmak:
This task has likely evolved to just:

- [x] Figure out how Firebase can be set to use VPN/https, set that, and then also set up access limitations (investigated in #96 )

holtzmak:
Firebase Database is only used over an encrypted connection - HTTPS is required (read [here](https://firebase.google.com/support/privacy#data_encryption) and [here for REST API](https://firebase.google.com/docs/reference/rest/database#:~:text=HTTPS%20is%20required.,create%20a%20Realtime%20Database%20instance.)). Database calls using/from Flutter application therefore must be using HTTPS, despite methods not explicitly stating HTTPs in the codelab (likely because doing so is redundant) (read [here](https://codelabs.developers.google.com/codelabs/flutter-firebase/index.html#10)).

[...]

PR #117, opened 2021-01-20
Setup firebase ios and android

MummenRider:
iOS and android are setup in firebase console.


holtzmak (changes requested):
I think the XCBuild files should not be checked in, the contain information specific to file locations on your machine. For example:
`"/Users/mumenrider/Projects/Humane-Transport/app/build/ios": {"is-mutated":true}`

These should be .gitignored by default in the `app/ios` gitignore.

- [x] Remove the XCBuilds and related binaries from the PR (git ignore them)

EDIT: I think the `build` folder is all that needs to be ignored, I'm not sure how it wasn't ignored on your machine

EDIT EDIT: I suppose we don't ignore the build file under `ios` directory, I suppose the build files appeared because you built from XCode - so, add `build/` to the `app/ios/.gitignore`

holtzmak:
I've de-linked this PR from #80 as we don't make a VPN connection yet, this is just configuration files?

holtzmak (approved):

PR #118, opened 2021-01-20
Added Sidebar and firebase configuration files

MummenRider:


holtzmak (approved):
Yeah, those google service files don't appear to be secretive. Thanks for fixing!

PR #120, opened 2021-01-22
Basic Log in auth

### LINKED ISSUE #94: Investigate and experiment with login authentication
Blocked by #96

Investigate and experiment with how we can use Shared Preferences/Firebase Authentication from #74 to save a logged in session on the mobile device.

This task is done when:
- [ ] We can use one of the above methods to save a logged in session, for some amount of time you deem satisfactory
- [ ] If you notice anything else useful, make record of that too (what else can we use the above for besides login?)

MummenRider:
Things I did for now:
- Manually added user in firebase to test log in
- Added bunch of TODOs for future log in auth implementation
- To pass integration test, I simply added the drawer.
- To sign in: email is clark@gmail.com and pass is capstone

### DIFF
--- a/app/lib/common/enums/app_state.dart
+++ b/app/lib/common/enums/app_state.dart
@@ -1,0 +1,1 @@
+enum ViewState { Idle, Busy }
--- a/app/lib/humane_transport_app.dart
+++ b/app/lib/humane_transport_app.dart
@@ -2,7 +2,6 @@ import 'package:app/common/style/style.dart';
 import 'package:app/providers/authentication/authentication.dart';
 import 'package:app/providers/navigation/navigation_provider.dart';
 import 'package:app/routes/main_navigator/app_route.dart';
-import 'package:app/screens/authenticate/auth_wrapper.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
@@ -23,7 +22,6 @@ class HumaneTransportApp extends StatelessWidget {
       child: Builder(
           builder: (context) => MaterialApp(
                 debugShowCheckedModeBanner: false,
-                initialRoute: AuthenticateWrapper.route,
                 onGenerateRoute: AppRouteGenerator.onGenerateRoute,
                 theme: ThemeData(
                   appBarTheme: AppBarTheme(color: MainAppColor),
--- a/app/lib/humane_transport_app.dart
+++ b/app/lib/humane_transport_app.dart
@@ -2,6 +2,7 @@ import 'package:app/common/style/style.dart';
 import 'package:app/providers/authentication/authentication.dart';
 import 'package:app/providers/navigation/navigation_provider.dart';
 import 'package:app/routes/main_navigator/app_route.dart';
+import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
@@ -13,7 +14,7 @@ class HumaneTransportApp extends StatelessWidget {
         ChangeNotifierProvider<NavigationProvider>(
             create: (context) => NavigationProvider()),
         ChangeNotifierProvider<AuthenticationProvider>(
-          create: (context) => AuthenticationProvider(),
+          create: (context) => AuthenticationProvider(FirebaseAuth.instance),
         ),
         StreamProvider(
             create: (context) =>
--- a/app/lib/humane_transport_app.dart
+++ b/app/lib/humane_transport_app.dart
@@ -1,6 +1,8 @@
 import 'package:app/common/style/style.dart';
-import 'package:app/providers/navigation_provider.dart';
+import 'package:app/providers/authentication/authentication.dart';
+import 'package:app/providers/navigation/navigation_provider.dart';
 import 'package:app/routes/main_navigator/app_route.dart';
+import 'package:app/screens/authenticate/auth_wrapper.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
@@ -11,10 +13,17 @@ class HumaneTransportApp extends StatelessWidget {
       providers: [
         ChangeNotifierProvider<NavigationProvider>(
             create: (context) => NavigationProvider()),
+        ChangeNotifierProvider<AuthenticationProvider>(
+          create: (context) => AuthenticationProvider(),
+        ),
+        StreamProvider(
+            create: (context) =>
+                context.read<AuthenticationProvider>().authStateChanges),
       ],
       child: Builder(
           builder: (context) => MaterialApp(
                 debugShowCheckedModeBanner: false,
+                initialRoute: AuthenticateWrapper.route,
                 onGenerateRoute: AppRouteGenerator.onGenerateRoute,
                 theme: ThemeData(
                   appBarTheme: AppBarTheme(color: MainAppColor),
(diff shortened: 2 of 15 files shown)


app/lib/humane_transport_app.dart
```diff
@@ -11,6 +12,12 @@ class HumaneTransportApp extends StatelessWidget {
       providers: [
         ChangeNotifierProvider<NavigationProvider>(
             create: (context) => NavigationProvider()),
+        ChangeNotifierProvider<AuthenticationProvider>(
+          create: (context) => AuthenticationProvider(),
+        ),
+        StreamProvider(
```
holtzmak, line 19: Why do we need a stream provider on top?

app/lib/main.dart
```diff
@@ -1,4 +1,9 @@
+import 'package:firebase_core/firebase_core.dart';
 import 'package:flutter/material.dart';
 import 'humane_transport_app.dart';
 
-void main() => runApp(HumaneTransportApp());
+Future<void> main() async {
```
holtzmak, line 5: I've not heard of having an `async` main. What does `Firebase.initializeApp()` do that we need it, and;
- [x] does it wipe user data on close? We want to cache ATRs.

app/lib/providers/authentication/authentication.dart
```diff
@@ -0,0 +1,58 @@
+import 'package:app/common/enums/app_state.dart';
+import 'package:firebase_auth/firebase_auth.dart';
+import 'package:flutter/material.dart';
+
+class AuthenticationProvider extends ChangeNotifier {
+  FirebaseAuth _firebaseAuth = FirebaseAuth.instance;
```
holtzmak, line 7: - [x] Are you sure this is a deep copy? That we're not making another instance?

app/lib/providers/authentication/authentication.dart
```diff
@@ -0,0 +1,58 @@
...
+  ViewState _viewState;
+
+  ViewState get viewState => _viewState;
+
+  void setViewState(ViewState viewState) {
+    _viewState = viewState;
+    notifyListeners();
+  }
+
+  Stream<User> get authStateChanges => _firebaseAuth.authStateChanges();
```
holtzmak, line 17: I think I see what `StreamProvider` does here, providing two-way binding. 
- [x] Why do we need both?

app/lib/providers/authentication/authentication.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  Future<void> signIn({String email, String password}) async {
+    try {
+      setViewState(ViewState.Busy);
+      await Future.delayed(Duration(seconds: 8));
```
holtzmak: - [x] Haha, is this leftovers?

app/lib/providers/authentication/authentication.dart
```diff
@@ -0,0 +1,58 @@
...
+      // TODO: Update UI with the error message
+    }
+  }
+
+  void signOut() async {
+    // TODO: Should handle exception error
+    await _firebaseAuth.signOut();
+  }
+
+  // TODO: Update account information
```
holtzmak, line 56: If you've created the issue/task already, we should reference it in these TODOs so we don't have floating work somewhere

app/test/main_test.dart
```diff
@@ -1,11 +1,11 @@
...
 
-void main() {
-  testWidgets('HumaneTransportApp says Go to page with bottom nav',
-      (WidgetTester tester) async {
-    await tester.pumpWidget(HumaneTransportApp());
-    final textFinder = find.text('Go to page with bottom nav');
-    expect(textFinder, findsOneWidget);
-  });
-}
+// void main() {
```
holtzmak: You should do what's called "skipping" the test (called disabling, ignoring, etc. in other test libraries). You can see the build was broken because we use `flutter test` but have no tests.

- [x] Skip the test, [instructions](https://pub.dev/packages/test#skipping-tests)

holtzmak (changes requested):
Some questions and requested changes for the broken test.

app/lib/main.dart
MummenRider: This is required to initialize Firebase to Flutter. You won't be able to use firebase without doing so

app/test/main_test.dart
MummenRider: Made changes here and finally able to pass the test. I added another screen called welcome (based on our 2nd design) and the test finally able to find the text

app/lib/humane_transport_app.dart
MummenRider: If it's not on top, then widgets above won't be able to listen to auth changes changes. The provider must and should always be above the widget who needs it. In this case, our Auth wrapper needs it.

app/lib/providers/authentication/authentication.dart
MummenRider: We need an instance of firebase in order to use its method like signInWithEmailAndPassword. We won't be able to use these firebase method without creating an instance of firebaseauth. Unless you know other method, please let me know.

app/lib/providers/authentication/authentication.dart
MummenRider: The on the top is like initialization and this one is the implementation. Which will listen to auth changes like user sign in or sign out. This is the one that will decide which screen user should go to.

app/lib/providers/authentication/authentication.dart
MummenRider: No issues/tasks created at the moment. I will link once issues are created.

app/lib/providers/authentication/authentication.dart
MummenRider: I removed this in my second commit

app/lib/main.dart
holtzmak: What does this mean? Does the Firebase "store" on the phone disconnect from the app when the app is closed?

- [ ] We should do something if we can't connect

app/lib/humane_transport_app.dart
holtzmak: What do you mean by widgets above?

app/lib/providers/authentication/authentication.dart
holtzmak: I thought FirebaseAuth.instance was a static item, but [these examples](https://firebase.flutter.dev/docs/auth/usage/) do what you do. No changes needed.

app/lib/common/enums/app_state.dart
```diff
@@ -0,0 +1 @@
+enum ViewState { Idle, Busy }
```
holtzmak, line 1: My open PR may break this... my problem

holtzmak (approved):
LGTM, just some questions (can merge and still answer questions on closed PR)

app/lib/humane_transport_app.dart
MummenRider: Hard to explain without some kind of widget tree diagram to show this. But I believe stream works similar to [provider](https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple). The article can explain it better and also shows a widget diagram

app/lib/main.dart
MummenRider: There has to be another way to initialize firebase to our app, but at the moment the tutorial I was following uses this method. At this point, I can't really find what's the best practice for initializing firebase. Everyone seems to have their own approach. If we use this approach, then you are correct! We should definitely do something if user can't connect.

app/lib/common/enums/app_state.dart
MummenRider: @holtzmak I tried  the code on that article and it worked perfectly. We don't even have use for ViewState anymore and can be removed.

PR #122, opened 2021-01-23
Add loading screen

### LINKED ISSUE #99: Experiment/Investigate using a simple loading screen

holtzmak:
Closes #99 .

Added a reusable loading screen. Credit to David at https://www.greycastle.se/loading-overlay-in-flutter/ for such an excellent [factory](https://flutterbyexample.com/lesson/factory-methods). I simply could not do better: this exact pattern shows how we can separate widget logic from business logic, among other very good practices. I stumbled across this example when I was investigating `barrierDismissible`.

I did the below on android because of the availability of a back button.

https://user-images.githubusercontent.com/32527219/105563916-872be880-5ce5-11eb-8349-08af16d5c518.mov

### DIFF
--- a/app/lib/common/widgets/loading_overlay_screen.dart
+++ b/app/lib/common/widgets/loading_overlay_screen.dart
@@ -1,0 +1,40 @@
+import 'package:app/common/style/style.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+
+/// A factory for a loading overlay screen. Credit https://www.greycastle.se/loading-overlay-in-flutter/
+class LoadingOverlayScreen {
+  BuildContext _context;
+
+  LoadingOverlayScreen._create(this._context);
+
+  factory LoadingOverlayScreen.of(BuildContext context) {
+    return LoadingOverlayScreen._create(context);
+  }
+
+  void hide() {
+    Navigator.of(_context, rootNavigator: true).pop();
+  }
+
+  void show() {
+    showDialog(
+        context: _context,
+        barrierDismissible: false,
+        child: WillPopScope(
+            child: _LoadingOverlayScreen(), onWillPop: () async => false));
+  }
+
+  Future<T> during<T>(Future<T> future) async {
+    show();
+    return future.whenComplete(() => hide());
+  }
+}
+
+class _LoadingOverlayScreen extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+        decoration: BoxDecoration(color: OverlayColor),
+        child: Center(child: CircularProgressIndicator()));
+  }
+}
--- a/app/lib/providers/authentication/authentication.dart
+++ b/app/lib/providers/authentication/authentication.dart
@@ -1,34 +1,20 @@
-import 'package:app/common/enums/app_state.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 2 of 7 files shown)


app/lib/common/widgets/loading_overlay_screen.dart
```diff
@@ -0,0 +1,40 @@
...
+
+  void hide() {
+    Navigator.of(_context, rootNavigator: true).pop();
+  }
+
+  void show() {
+    showDialog(
+        context: _context,
+        barrierDismissible: false,
+        child: WillPopScope(
```
holtzmak, line 23: Since we have two navigators, we have to be careful not to allow the user to back out of the loading screen and mess up our history. The absence of events when attempting to click or interact with the underlying screens means `barrierDismissible` works. We must use it alongside the `WillPopScrope` to prevent navigation interaction at the same time.

app/lib/common/widgets/loading_overlay_screen.dart
```diff
@@ -0,0 +1,40 @@
...
+            child: _LoadingOverlayScreen(), onWillPop: () async => false));
```
holtzmak, line 24: [Interesting way of providing onWillPop](https://stackoverflow.com/questions/53156666/detect-back-button-press-while-dialog-is-open-in-flutter)

MummenRider (approved):
LGTM, I tried it in my code and worked as expected. I also noticed that I don't have to keep track of view changes when I used LoadingOverlayScreen widget. What we can do in the future:
- [x] Remove the ViewState in AuthenticationProvider

holtzmak:
I agree - it is better to have futures drive the loading screen than to set the state of the underlying widget. In a sense though, we still have the `ViewState`, it's the show/hide methods tied to the future's status.

app/lib/providers/authentication/authentication.dart
```diff
@@ -1,44 +1,27 @@
...
-    notifyListeners();
-  }
+  FirebaseAuth _firebaseAuth = FirebaseAuth.instance;
 
   Stream<User> get authStateChanges => _firebaseAuth.authStateChanges();
 
   Future<void> signIn({String email, String password}) async {
     try {
-      setViewState(ViewState.Busy);
-      await Future.delayed(Duration(seconds: 2));
```
holtzmak, line 22: Dumped these, no need to wait

app/lib/screens/authenticate/sign_in.dart
```diff
@@ -31,18 +32,14 @@ class SignInScreen extends StatelessWidget {
...
-                        onPressed: () {
-                          model.signIn(
-                              email: emailController.text.trim(),
-                              password: passwordController.text.trim());
-                        },
-                        child: Text('Sign In'),
-                        color: Colors.green,
-                      ),
+                RaisedButton(
+                  onPressed: () async => LoadingOverlayScreen.of(context)
```
holtzmak, line 36: Sign in behaviour has not changed, once the work is done the loading overlay screen is popped and the new screen navigated to without conflict.

PR #123, opened 2021-01-25
add upload-image feature

mansiPaTeLL:
added upload-image feature

### DIFF
--- a/app/ios/Runner/Info.plist
+++ b/app/ios/Runner/Info.plist
@@ -41,5 +41,11 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>NSCameraUsageDescription</key>
+    <string>Can I use the camera please?</string>
+    <key>NSMicrophoneUsageDescription</key>
+    <string>Can I use the mic please?</string>
+    <key>NSPhotoLibraryUsageDescription</key>
+    <string>Example usage description</string>
 </dict>
 </plist>
--- a/app/lib/common/utilities/image_upload.dart
+++ b/app/lib/common/utilities/image_upload.dart
@@ -1,0 +1,215 @@
+import 'dart:io';
+import 'dart:async';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:image_picker/image_picker.dart';
+
+class UploadImage extends StatefulWidget {
+  const UploadImage({Key key}) : super(key: key);
+  static const route = '/image_upload';
+  @override
+  _UploadImageState createState() => _UploadImageState();
+}
+
+class _UploadImageState extends State<UploadImage> {
+  PickedFile _imageFile;
+  dynamic _pickImageError;
+  File _image;
+  final ImagePicker _picker = ImagePicker();
+  final TextEditingController maxWidthController = TextEditingController();
+  final TextEditingController maxHeightController = TextEditingController();
+  final TextEditingController qualityController = TextEditingController();
+
+  void _onImageButtonPressed(ImageSource source, {BuildContext context}) async {
+    try {
+      final pickedFile = await _picker.getImage(
+        source: source,
+      );
+      setState(() {
+        _imageFile = pickedFile;
+      });
+    } catch (e) {
+      setState(() {
+        _pickImageError = e;
+      });
+    }
+  }
+
+  Widget _previewImage() {
+    //final Text retrieveError = _getRetrieveErrorWidget();
+    if (_imageFile != null) {
+      if (kIsWeb) {
+        // Why network?
+        // See https://pub.dev/packages/image_picker#getting-ready-for-the-web-platform
+        return Image.network(_imageFile.path);
+      } else {
+        return Semantics(
+            child: Image.file(File(_imageFile.path)),
+            label: 'image_picker_example_picked_image');
+      }
+    } else if (_pickImageError != null) {
+      return Text(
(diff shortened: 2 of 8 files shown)


app/lib/common/utilities/image_upload.dart
```diff
@@ -0,0 +1,218 @@
...
+      });
+    } catch (e) {
+      setState(() {
+        _pickImageError = e;
+      });
+    }
+  }
+
+  Widget _previewImage() {
+    //final Text retrieveError = _getRetrieveErrorWidget();
```
holtzmak: - [x] Is this meant to be there or can it be removed?

app/lib/common/utilities/image_upload.dart
```diff
@@ -0,0 +1,218 @@
...
+      floatingActionButton: Column(
+        mainAxisAlignment: MainAxisAlignment.end,
+        children: <Widget>[
+          GestureDetector(
+            onTap: () {
+              _showPicker(context);
+            },
+            child: CircleAvatar(
+              radius: 55,
+              backgroundColor: Color(0xffFDCF09),
```
holtzmak: - [x] Move color to `style.dart` if you think we can reuse it

app/lib/common/utilities/image_upload.dart
```diff
@@ -0,0 +1,218 @@
...
+          return SafeArea(
+            child: Container(
+              child: new Wrap(
+                children: <Widget>[
+                  new ListTile(
+                      leading: new Icon(Icons.photo_library),
+                      title: new Text('Photo Library'),
+                      onTap: () {
+                        _imgFromGallery();
+                        Navigator.of(context).pop();
```
holtzmak, line 181: Which navigator is this? I'm not sure we want to pop the one with the bottom nav nor would we want the bottom nav when in the photo gallery

app/lib/routes/inner_navigator/new_travel_route.dart
```diff
@@ -14,6 +15,8 @@ class NewTravelRouteGenerator {
             return TestScreenOne();
           case TestScreenTwo.route:
             return TestScreenTwo();
+          case UploadImage.route:
```
holtzmak, line 18: @MummenRider is it a good idea to dedicate a screen for images/camera? Would we want it to be an overlay without bottom nav?

holtzmak (changes requested):
Are we keeping the camera stuff? I thought you mentioned in standup @mansiPaTeLL that the camera was buggy and we should just use image picker from gallery.

- [x] Remove the camera plugins and camera-specific code if we're not using them

app/lib/common/utilities/image_upload.dart
```diff
@@ -0,0 +1,218 @@
+import 'dart:io';
```
holtzmak, line 1: - [x] Additionally, I would like to see this file split up. Too big.

mansiPaTeLL:
> Are we keeping the camera stuff? I thought you mentioned in standup @mansiPaTeLL that the camera was buggy and we should just use image picker from gallery.
> 
>     * [ ]  Remove the camera plugins and camera-specific code if we're not using them

@holtzmak image_picker can do both photo-library and image pick by camera. So it is good to have here. There was a different plugin "camera" was not working properly. No need to remove this.

app/lib/common/utilities/image_upload.dart
mansiPaTeLL: @holtzmak This navigator is for when we are in photo gallery and after selecting image, we want to go back to our app. So, `Navigator.of(context).pop();` can be used. Because we don't want to have bottom nav when in photo gallery. So we don't want to use route here as we are in different app.

app/lib/routes/inner_navigator/new_travel_route.dart
MummenRider: It would look weird  if there's a bottom nav on the camera page. For now, we can use what we have here since this is just experimentation.

app/lib/common/utilities/image_upload.dart
mansiPaTeLL: @holtzmak Since, this is an experimental task, I think we should keep every components related to image in here. Later when we are actually building the GUI, we can definitely take each component wherever we need.

app/lib/common/utilities/image_upload.dart
mansiPaTeLL: @holtzmak Thanks Kelly. I removed it :)

app/lib/common/utilities/image_upload.dart
mansiPaTeLL: @holtzmak Added Color to style.dart. This `CircleAvatar` we will only use at profile picture.

app/ios/Runner/Info.plist
```diff
@@ -41,5 +41,11 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>NSCameraUsageDescription</key>
+    <string>Can I use the camera please?</string>
+    <key>NSMicrophoneUsageDescription</key>
+    <string>Can I use the mic please?</string>
+    <key>NSPhotoLibraryUsageDescription</key>
+    <string>Example usage description</string>
```
holtzmak, line 49: Should this says something like "Can I use the Photo Library"?

holtzmak (approved):
LGTM for now. I will want lots of refactoring when we go to use this right in the future. Minor change requested for one of the request messages as it still says `Example message here`.

PR #136, opened 2021-01-28
Create AnimalTransportRecord object

### LINKED ISSUE #129: Create the Animal Transport Record object
- [x] Using our consolidated ATR in `documentation/specification`, create our code object(s) that represent the form in the application

This task blocks most other work, so it should be done first.

Be mindful of a few things:
1. That the object(s) are scalable (for example, FWR list so that we may programmatically add more records)
2. That the object(s) can be replaced easily
3. That the object(s) have some relation to our Firebase database data structure for ease of understanding

_Consider further the convention for objects like this when filling out forms in a Flutter application._

holtzmak:
Closes #129 .

I referenced our database architecture drawings and our consolidated ATR under `documentation`. I did not add unit tests for these object classes as they contain no logic. `AnimalTransportRecord` is a curious case of an immutable class with creators, but these are simply constructors and unit tests for each function would be excessive.

Note that the creator functions use `this`, whereas a factory function cannot - hence why ATR creators are not factories.

### DIFF
--- a/app/lib/common/utilities/address.dart
+++ b/app/lib/common/utilities/address.dart
@@ -1,0 +1,13 @@
+import 'package:flutter/material.dart';
+
+@immutable
+class Address {
+  final String streetAddress;
+  final String city;
+  final String provinceOrState;
+  final String country;
+  final String postalCode;
+
+  Address(this.streetAddress, this.city, this.provinceOrState, this.country,
+      this.postalCode);
+}
--- a/app/lib/common/utilities/address.dart
+++ b/app/lib/common/utilities/address.dart
@@ -8,6 +8,15 @@ class Address {
   final String country;
   final String postalCode;
 
-  Address(this.streetAddress, this.city, this.provinceOrState, this.country,
-      this.postalCode);
+  Address(
+      {@required String streetAddress,
+      @required String city,
+      @required String provinceOrState,
+      @required String country,
+      @required String postalCode})
+      : streetAddress = streetAddress,
+        city = city,
+        provinceOrState = provinceOrState,
+        country = country,
+        postalCode = postalCode;
 }
--- a/app/lib/common/utilities/optional.dart
+++ b/app/lib/common/utilities/optional.dart
@@ -1,0 +1,14 @@
+import 'package:flutter/foundation.dart';
+
+/// Inspired by Java's Optional type.
+@immutable
+class Optional<T> {
+  final T _optional;
+
+  Optional(this._optional);
+
+  bool isPresent() => _optional != null;
+
+  T get() =>
+      isPresent() ? _optional : throw Exception("The Optional value is null!");
+}
--- a/app/lib/models/acknowledgement_info.dart
+++ b/app/lib/models/acknowledgement_info.dart
@@ -1,0 +1,12 @@
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+
+@immutable
+class AcknowledgementInfo {
+  final File shipperAck;
+  final File transporterAck;
+  final File receiverAck;
+
+  AcknowledgementInfo(this.shipperAck, this.transporterAck, this.receiverAck);
+}
--- a/app/lib/models/acknowledgement_info.dart
+++ b/app/lib/models/acknowledgement_info.dart
@@ -8,5 +8,11 @@ class AcknowledgementInfo {
   final File transporterAck;
   final File receiverAck;
 
-  AcknowledgementInfo(this.shipperAck, this.transporterAck, this.receiverAck);
+  AcknowledgementInfo(
+      {@required File shipperAck,
+      @required File transporterAck,
+      @required File receiverAck})
+      : shipperAck = shipperAck,
+        transporterAck = transporterAck,
+        receiverAck = receiverAck;
 }
--- a/app/lib/models/animal_transport_record.dart
+++ b/app/lib/models/animal_transport_record.dart
@@ -1,0 +1,52 @@
+import 'package:app/models/shipper_info.dart';
+import 'package:app/models/transporter_info.dart';
+import 'package:flutter/material.dart';
+
+import 'acknowledgement_info.dart';
+import 'contingency_plan_info.dart';
+import 'delivery_info.dart';
+import 'feed_water_rest_info.dart';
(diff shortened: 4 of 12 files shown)


app/lib/common/utilities/address.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/material.dart';
+
+@immutable
+class Address {
+  final String streetAddress;
+  final String city;
+  final String provinceOrState;
+  final String country;
```
holtzmak, line 8: I imagine the String type to be replaced with enum in the future for robustness, depends if we use a location picker or something in the ATR editing screen.

app/lib/common/utilities/optional.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:flutter/foundation.dart';
+
+/// Inspired by Java's Optional type.
```
holtzmak, line 3: [Java's optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html). I don't have this for every type, not sure if overkill as empty string will suffice.

app/lib/models/acknowledgement_info.dart
```diff
@@ -0,0 +1,12 @@
+import 'dart:io';
+
+import 'package:flutter/material.dart';
+
+@immutable
+class AcknowledgementInfo {
+  final File shipperAck;
```
holtzmak, line 7: File type as we should take the person's signature. Can be replaced with whatever type our database uses for images.

app/lib/models/animal_transport_record.dart
```diff
@@ -0,0 +1,52 @@
...
+  final FeedWaterRestInfo fwrInfo;
+  final LoadingVehicleInfo vehicleInfo;
+  final DeliveryInfo deliveryInfo;
+  final AcknowledgementInfo ackInfo;
+  final ContingencyPlanInfo contingencyInfo;
+
+  AnimalTransportRecord(this.shipInfo, this.tranInfo, this.fwrInfo,
+      this.vehicleInfo, this.deliveryInfo, this.ackInfo, this.contingencyInfo);
+
+  AnimalTransportRecord withShipInfo(ShipperInfo newShipInfo) =>
```
holtzmak, line 37: This is a creator method (uses `this` fields) and not a factory, which is like a static creator.

app/lib/models/contingency_plan_info.dart
```diff
@@ -0,0 +1,74 @@
...
+      this.goalStatement,
+      this.communicationPlan,
+      this._crisisContacts,
+      this.expectedPrepProcess,
+      this.standardAnimalMonitoring,
+      this._potentialHazards,
+      this._potentialSafetyActions,
+      this._contingencyEvents);
+
+  List<String> crisisContacts() => List.unmodifiable(_crisisContacts);
```
holtzmak, line 32: Lists aren't immutable in dart, so we do this in immutable classes with lists.

app/lib/common/utilities/address.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/material.dart';
+
+@immutable
```
holtzmak, line 3: A nice little annotation that prevents people from accidentally forgetting `final` keyword.

app/lib/models/animal_transport_record.dart
holtzmak: This is good to have, as these methods will support splitting the ATR editing across screens, and also still support editing in once screen with accordion menus.

MummenRider (changes requested):
LGTM, few changes I'd like to see

- [x] If we know an object requires a specific parameter (e.g. postal code), then we should definitely add a @required meta tag. Though you have test that will throw error if parameter is missing, I think @required meta tag should be use as per best practice.
- [x] Wrap the constructor with curly braces as this give as ability to use named parameters. This is especially useful if an object have 3 or more parameters.

holtzmak:
@MummenRider every field of these classes is required, I think using the meta tag is overkill. The `Optional` type gives you the option to have a field as essentially `null` or not - from my investigation of the consolidated documents, just the one field is optional and that is the receiver's account.

MummenRider:
> @MummenRider every field of these classes is required, I think using the meta tag is overkill. The `Optional` type gives you the option to have a field as essentially `null` or not - from my investigation of the consolidated documents, just the one field is optional and that is the receiver's account.

Ohh... So if properties don't have Optional type, then that implies it's a required parameters? I see that the Optional type object will check if parameters is null, but is this true for all properties? Like if I created an instance contigencyplaninfo object and I missed 1 parameter, would this give me an error message (as it should be)?

holtzmak:
> Ohh... So if properties don't have Optional type, then that implies it's a required parameters? I see that the Optional type object will check if parameters is null, but is this true for all properties? Like if I created an instance contigencyplaninfo object and I missed 1 parameter, would this give me an error message (as it should be)?

Yes. You get this error if you miss a parameter. It won't compile.

Screen Shot 2021-01-28 at 9 23 57 AM

I can't stop you from putting `null` for a parameter, as dart [null-safety is in beta](https://dart.dev/null-safety) (which would enforce non-null parameters unless we add a `?` to the field). But, by convention we should not be putting null in any objects: if it must be null wrap it in a safe-fail object like `Optional` with method `isPresent()` (there's more methods we can use, I haven't added them yet).

MummenRider:
Can you try this in a Constructor wrapped with curly braces? Did you get the same error? I'm pretty sure you won't get an error. Please correct me I am wrong, but my knowledge is by doing named parameters, this also means everything inside the braces will be optional. To fix this, you need to add required meta tag. 
> Yes. You get this error if you miss a parameter. It won't compile.
> 
> Screen Shot 2021-01-28 at 9 23 57 AM

holtzmak:
> Can you try this in a Constructor wrapped with curly braces? Did you get the same error? I'm pretty sure you won't get an error. Please correct me I am wrong, but my knowledge is by doing named parameters, this also means everything inside the braces will be optional. To fix this, you need to add required meta tag.

You are correct, we will need the required annotation for such a case... Why can't everything be like kotlin 😢 .

MummenRider (approved):
Changes here LGTM

PR #137, opened 2021-01-28
Update iphone sim used to iOS 14.4

holtzmak:
Xcode updated in the last few days to iOS 14.4, this changes what simulators are available on GItHub machines. Likely, the down period last night was them updated their machines.


holtzmak:
No need for review on this, I will merge once check is done.

PR #138, opened 2021-01-28
Changed iOS to 14.3 (stable)

holtzmak:
Follow up to prior PR - XCode updated to 14.4, but GItHub updated to 14.3. We literally have the command "list all devices" in the build flutter action so I would know what the latest devices were, but I missed it. I put it there!!! >:(


holtzmak:
Again, no need for review.

PR #139, opened 2021-01-28
Email pdf

### LINKED ISSUE #97: Experiment with sending an email with PDF attached from Flutter
This task is a combination of #81 and #79.

This task is done when:
- [ ] You can send an email with a PDF attached
- [ ] You can send an email with many PDFs attached

SanaKhan1997:


app/android/app/src/debug/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.HumaneTransport">
+          package="com.HumaneTransport">
```
holtzmak, line 2: Do you still have your windows machine for development?

app/android/settings_aar.gradle
```diff
@@ -0,0 +1 @@
+include ':app'
```
holtzmak: Why do we have this?

app/lib/common/utilities/email_sender.dart
```diff
@@ -1,40 +1,47 @@
...
     try {
       await FlutterEmailSender.send(email);
-      emailResponse = 'success';
+      emailResponse = 'Email Sent';
     } catch (error) {
       emailResponse = error.toString();
     }
-
-    print(emailResponse);
+    return emailResponse;
```
holtzmak: I'm not sure you need to have the return type as `Future<String>` it can be just `String`. I think the `await` is blocking, so the function won't return until email response is set so it's not a true future.

The reason why what you have works is because dart is smart enough to wrap your string in a Future. This is useful for some things, but not futures.

- [x] Change return type to just `String`

app/lib/common/utilities/email_sender.dart
```diff
@@ -62,9 +69,29 @@ class EmailRoute extends StatelessWidget {
                   onPressed: () {
                     if (formKey.currentState.validate()) {
                       print('valid');
-                      email.send(bodyController.text, subjectController.text,
-                          recipientController.text);
+                      //email.send(bodyController.text, subjectController.text,recipientController.text, pdf);
```
holtzmak: - [x] Remove commented out code

app/lib/common/utilities/email_sender.dart
```diff
@@ -62,9 +69,29 @@ class EmailRoute extends StatelessWidget {
...
+                      Future.value(email.send(
```
holtzmak: As we saw earlier, `send` shouldn't have a future return type. You can probably just launch showDialog as asynchronous and it will appear when the function is done. If you want it blocking interaction, you'll need to use `barrierDismissible` like `loading_screen_overlay.dart`. 

- [x] The changes to the return type should force changes here

app/lib/common/utilities/email_sender.dart
```diff
@@ -62,9 +69,29 @@ class EmailRoute extends StatelessWidget {
...
+                                      onPressed: () {
+                                        Navigator.pop(context);
+                                      },
+                                      child: Text(email.emailResponse),
+                                    ),
+                                  );
+                                },
+                              ));
                     }
+                    Navigator.of(context).pushNamed(TravelHistory.route);
```
holtzmak: This seems out of place? Why is this here?

holtzmak: Are we to go back when clicking submit? This doesn't seem obvious

app/lib/common/utilities/pdf_preview.dart
```diff
@@ -34,17 +35,38 @@ class _PdfPreviewState extends State<PdfPreview> {
...
+                //If PDF scaffold state changed switch to email_sender
+                return EmailRoute(pdf: snapshot.requireData);
+              } else
                 return PDFViewerScaffold(
                   appBar: AppBar(
                     title: Text('Example PDF'),
+                    actions: [
+                      OutlinedButton.icon(
+                          onPressed: () {
+                            exitPdf();
```
holtzmak: I think this is anti-pattern, you should push the email PDF from navigator instead.

- [x] Push the email widget into tree with like `Navigator.of(context).pushNamed...` because you have indeed added the `route` to routes already

app/pubspec.lock
```diff
@@ -78,6 +78,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.1.5"
+  csslib:
```
holtzmak: CSS and html? I did not need this when attempting email with PDF.

- [x] Remove this if we don't need them

app/pubspec.yaml
```diff
@@ -28,6 +28,7 @@ dependencies:
   firebase_core: ^0.7.0
   flutter_email_sender: ^4.0.0
   flutter_full_pdf_viewer: ^1.0.6
+  html: ^0.14.0+4
```
holtzmak: - [x] Remove

holtzmak (changes requested):
Some changes I'd like to see: minor refactoring to make it more understandable and to use Futures correctly, and also to remove some excess things that snuck through.

app/lib/common/utilities/pdf_preview.dart
SanaKhan1997: I tried Navigator initially but it turns out we have to deactivate pdf scaffold and pass pdf file to EmailRoute

app/lib/common/utilities/email_sender.dart
SanaKhan1997: Yeah I added to see if we can return to travel history screen. My bad I forgot to remove it.

app/lib/common/utilities/email_sender.dart
SanaKhan1997: This is what I initially wanted but since send is async function it has to have Future as return type. Also when I changed it to String, dialog box occurs before gmail launch.

MummenRider (commented):
All issues/concern was already reviewed by Kelly. We know that email functionality works, but once we start coding for real, we'll have to remind ourselves to follow best practices. All in all, good job!

app/lib/common/utilities/pdf_preview.dart
```diff
@@ -38,6 +38,23 @@ class _PdfPreviewState extends State<PdfPreview> {
                 return PDFViewerScaffold(
                   appBar: AppBar(
                     title: Text('Example PDF'),
+                    actions: [
+                      OutlinedButton.icon(
+                          onPressed: () {
+                            Navigator.pop(context);
```
holtzmak, line 44: I think this is dangerous because this is popping the navigator while the widget is still live. We should refactor this int he future.

app/lib/common/utilities/email_sender.dart
```diff
@@ -1,40 +1,46 @@
+import 'dart:io';
+import 'package:flutter/cupertino.dart';
 import 'package:flutter_email_sender/flutter_email_sender.dart';
 import 'package:flutter/material.dart';
 
 class EmailSender {
-  Future<void> send(
-      String emailBody, String eSubject, String eRecipients) async {
+  String emailResponse;
```
holtzmak, line 7: Seems odd to have the email response be a field of the class, it should have been defined in `send` like before. Something to fix in the future.

app/pubspec.lock
```diff
@@ -36,20 +36,6 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.1.0-nullsafety.1"
-  camera:
```
holtzmak, line 39: Good change I think. Missed in Mansi's last PR

app/pubspec.lock
```diff
@@ -544,4 +530,4 @@ packages:
     version: "4.5.1"
 sdks:
   dart: ">=2.10.0-110 <2.11.0"
-  flutter: ">=1.22.0 <2.0.0"
+  flutter: ">=1.16.0 <2.0.0"
```
holtzmak, line 519: We had this problem with Mansi's PR yesterday.

- [x] Change back to 1.22.0

app/pubspec.lock
```diff
@@ -85,20 +71,20 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.1.1"
-  cross_file:
+  crypto:
```
holtzmak, line 74: You changes do not make sense. You've removed cross-file and replaced it with crypto. Do we even need this?
Additionally, you use csslib over crypto later. 

- [x] Remove all of these if we don't need them

app/pubspec.lock
```diff
@@ -444,13 +437,6 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "2.1.0-nullsafety.1"
-  stream_transform:
```
holtzmak, line 447: I thought @MummenRider was using this?

holtzmak (changes requested):
Make some changes because stuff got messed up in rebase, please.

app/pubspec.lock
MummenRider: Nope, I don't remember adding this. If you're talking about stream or streambuilder, then that is different
> I thought @MummenRider was using this?

app/pubspec.lock
holtzmak: Flutter pub upgrade made this change

app/pubspec.lock
holtzmak: flutter pub upgrade made these changes

app/pubspec.lock
holtzmak: Yup, flutter pub upgrade was doing something. I deleted pubspec.lock and ran flutter pub get again, and this is still gone. Likely some funkiness with rebases here.

holtzmak (approved):
Doubled-checked myself that pub changes were meant to be, and they are.

PR #140, opened 2021-01-28
Upload image - changed sdk

mansiPaTeLL:


holtzmak (approved):
Removing the camera plugin is correct. You will need to rebase this branch before merging.

Please note that some of the error messages don't look very good, we will need to swap these with `showDialogues` or something. You can do this now or later, up to you.

Screen Shot 2021-01-28 at 6 05 12 PM

Issue #130, opened 2021-01-26
Create the Firebase Database interface implementation


holtzmak:
Related to #70 , and blocked by it. Also blocked by #129. Also blocked by #124.

- [x] Create the Firebase interface implementation that implements the generic queries in #70 to communicate with our Firebase instance

There should be unit (or integration) tests for this code using the object in #129. The tests may communicate with a mocked database at this time so we don't go over our Spark plan limit.

*If you are mocking the database instance, then the interface implementation should have a Firebase instance injected into it - meaning it takes it through the constructor!*

SanaKhan1997:
When implementing atr queries, create two methods for retrieving bunch of documents:

 - [ ]  Method which only retrieves latest 10(use limit and order) atr documents
 - [ ]  Method which retrieves all documents

PR #142, opened 2021-01-30
Re-strcuture files

### LINKED ISSUE #143: Re-structure files based on MVVM architecture
Files are organized based on MVVM

MummenRider:
This commit's purpose is to restructure files based on [MVVM architecture](https://www.raywenderlich.com/6373413-state-management-with-provider#toc-anchor-016).
Note: The overlay screen was created by [FilledStack](https://www.filledstacks.com/). They're also popularized MVVM to Flutter community. 
Here are some important changes:
- Get it plugin is added and is currently implemented in login authentication. By using get it, we can 💯  say that our UI and business logic is indeed separated. To reviewers, I did not use get it to all of our services because it will take some time to do. But mainly because I'm still noob. I implemented get it only in initial part of our app hoping this will serve as a template once we code for real.
- Everything that is related to UI are moved to `ui` directory
- Authentication Service will contain all things that are related to firebase authentication such as sign in and log in methods. Due to MVVM approach, this do not communicate with our UI stuff. To communicate with our UI, we will use our view models.
- Navigation Service will serve as a bridge to navigate screens. Due to complete separation of UI and our business logic, we do not have access to `context` anymore. Usually we do `Navigator.of(context, rootNavigator: true).pushNamed(TestScreenOne.route)`, doing this will violate the MVVM architecture. The UI shouldn't have any ability to dictate what should happen in our application. Our business logic will do that.
- I did not use our original` loading overlay screen` widget to display the awaiting period because it violates the MVVM. The article I followed shows how to properly to apply overlay screen which follows MVVM. One thing I really like about this is it's highly customizable!
- ChangeNotifierProvider wraps all of our UI's and is located to our BaseView. This is exactly the same when it was located to HumaneTransportApp, but only tree below. You don't have to touch this. 
- We also have BaseViewModel which will contain similar aspects of the view model like current state of the screen.

Some important rules I currently know in order to use MVVM:
- UI should never use service to change anything, this apply to navigation.
- Each screen/view must have their own view model in order to communicate with our data. So if you have screen like sign in that communicates with a data, then you need to create view model
Screen Shot 2021-01-29 at 9 19 47 PM

Our current overlay

https://user-images.githubusercontent.com/43891568/106345787-40576900-6278-11eb-899a-7438c98b6cfb.mov


app/lib/core/services/navigation/nav_service.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:flutter/material.dart';
+
+class NavigationService {
```
holtzmak, line 3: I like this change, abstracting out how we navigate should increase understandability and reduce our need to describe why we have/had two navigators

app/lib/core/services/service_locator.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'navigation/nav_service.dart';
+
+final locator = GetIt.instance;
+
+void setUpLocator() {
+  locator.registerLazySingleton<AuthenticationService>(
+      () => AuthenticationService());
+  locator.registerLazySingleton<NavigationService>(() => NavigationService());
+
+  locator.registerFactory<WelcomeViewModel>(() => WelcomeViewModel());
```
holtzmak, line 16: Hm, is this the intended use of direct injection for the models? This strikes me as excessive, but I see the usefulness of it as we should technically be able to launch widgets any where we need.

holtzmak: We could just have factories in the widgets which I believe is the intended use (like what `loading_overlay_screen.dart` had). They are static and available anywhere.

app/lib/core/services/service_locator.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'package:app/core/view_models/signout_view_model.dart';
+import 'package:app/core/view_models/welcome_view_model.dart';
+import 'package:get_it/get_it.dart';
+import 'authentication/auth_service.dart';
+import 'navigation/nav_service.dart';
+
+final locator = GetIt.instance;
+
+void setUpLocator() {
+  locator.registerLazySingleton<AuthenticationService>(
```
holtzmak, line 12: It makes sense to have services as singletons, but not widget factories as a singleton registered in `GetIt`. At least, to me.

app/lib/core/view_models/signin_view_model.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:app/core/services/authentication/auth_service.dart';
+import 'package:app/core/services/navigation/nav_service.dart';
+import 'package:app/core/services/service_locator.dart';
+import 'package:app/core/view_models/base_view_model.dart';
+import 'package:app/ui/views/home/home.dart';
+import 'package:flutter/material.dart';
+
+class SignInViewModel extends BaseViewModel {
+  //Inject auth service
+  final AuthenticationService _authenticationService =
```
holtzmak, line 11: Again, these services as singletons make some sense because we may need the service across the app (and we don't want to hand around our single service nor do we want to be spinning up new instances all the time - unlike the widgets where we should have some static factory but we don't need or want the same instances all the time)

app/lib/core/view_models/signout_view_model.dart
```diff
@@ -0,0 +1,22 @@
...
+  //Inject auth service
+  final AuthenticationService _authenticationService =
+      locator<AuthenticationService>();
+  final NavigationService _navigationService = locator<NavigationService>();
+
+  Future signOut() async {
+    // TODO: Should handle exception error
+    try {
+      await _authenticationService.signOut();
+      _navigationService.pop();
```
holtzmak, line 15: The asynchronous code allows the pop before print, but I think we really should pop after the print (make navigation last) because traditionally, we would have exited the stack call at pop which would have left this function hanging. Recall the jumps in ENSE 352.

This is working because the UI is separate from executing code (popping is like calling hide), but every time I see this I think it smells.

app/lib/ui/views/signin/sign_in.dart
```diff
@@ -0,0 +1,69 @@
...
+                    key: Key('emailKey'),
+                    decoration: InputDecoration(hintText: "Email"),
+                    controller: emailController,
+                  ),
+                  TextFormField(
+                    controller: passwordController,
+                    decoration: InputDecoration(hintText: "Password"),
+                  ),
+                  RaisedButton(
+                    // onPressed: () async => LoadingOverlayScreen.of(context)
```
holtzmak: - [x] Remove this if we are no longer using it. Class can get booted too if we are using `BusyOverlay`.

app/lib/ui/views/welcome/welcome.dart
```diff
@@ -0,0 +1,40 @@
...
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: [
+              Text(
+                "Welcome",
+                key: Key('welcome'),
+              ),
+              // TODO: Register button
+              RaisedButton(
+                onPressed: () {
+                  Navigator.of(context).pushNamed(SignInScreen.route);
```
holtzmak, line 25: - [x] Should we not be using the service here?

app/lib/ui/widgets/list_tile.dart
```diff
@@ -30,7 +30,7 @@ class TemplateListTile extends StatelessWidget {
       child: InkWell(
         onTap: () {
           Navigator.of(context).pop();
```
holtzmak, line 32: - [x] I thought we were no longer using context?

I could definitely be wrong. Does this work since we give our main the navigator service as its context?

app/pubspec.lock
```diff
@@ -36,6 +50,41 @@ packages:
...
+    source: hosted
+    version: "4.3.2"
+  built_value:
+    dependency: transitive
+    description:
+      name: built_value
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "7.1.0"
+  camera:
```
holtzmak: This aren't mean to be here!!!

- [x] Delete `pubspec.lock` and run `flutter pub get/ upgrade` to add it back

app/pubspec.lock
```diff
@@ -514,6 +668,13 @@ packages:
...
+    description:
+      name: yaml
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "2.2.1"
 sdks:
-  dart: ">=2.10.0-110 <2.11.0"
-  flutter: ">=1.16.0 <2.0.0"
+  dart: ">=2.10.0 <2.11.0"
+  flutter: ">=1.22.0 <2.0.0"
```
holtzmak: The minimum should definitely be 1.16, unless your newest plugin changed this. I think not though, because you have the camera dependencies above which we aren't using anymore.

app/pubspec.yaml
```diff
@@ -23,15 +23,18 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  camera: ^0.7.0
```
holtzmak: - [x] Remove this!

app/test/main_test.dart
```diff
@@ -1,10 +1,18 @@
...
 
 void main() {
-  testWidgets('HumaneTransportApp says Welcome', (WidgetTester tester) async {
-    await tester.pumpWidget(HumaneTransportApp());
-    final textFinder = find.text('Welcome');
-    expect(textFinder, findsOneWidget);
+  group('Group test', () {
+    // Needed to show where the implementations are located
+    setUpAll(() {
+      setUpLocator();
```
holtzmak, line 9: We'll likely have a test version of this which implements the locator. We'll shove mocks into it.

- [x] Might as well make our setUpLocator() mockable -> make it so main gives it things like AuthenticationService and such (give it a constructor)

We will end up having two mains: our main-main which is dumb and gives static things to GetIt, and our testing main which takes in dependencies and sets up it's getIt with them. This way the we can make all app dependencies mockable at once!!

holtzmak (changes requested):
Looks like some pubspec-y things got messed up, but easy to fix. I have some questions about how you're using our direct injection service - the widgets in the injector seem odd to me.

Other minor questions and suggestions here and there. I like the changes I'm seeing.

app/lib/core/services/service_locator.dart
MummenRider: Yes, every viewmodels and services are required to be added in the locator. Without this our app wont be able to communicate data to UI. As for the type, I can remove that. Not really necessary, this is similar to <Widgets> child:

app/lib/core/services/service_locator.dart
MummenRider: trust me, this is needed. It wont work otherwise

app/lib/core/view_models/signout_view_model.dart
MummenRider: Will switch this  two in my next commit

app/pubspec.lock
MummenRider: weird that it's here. Noted and will remove it

app/test/main_test.dart
MummenRider: I plan to this next time. I already have task for API in my queue. I will do this after API task. I also added mockito already as we're going use it in the near future

app/lib/ui/widgets/list_tile.dart
MummenRider: I haven't implemented any kind of services/viewmodels here as I already mentioned

app/lib/ui/views/welcome/welcome.dart
MummenRider: Ideally, yes, we should be using services across the application. But as I already mentioned above, refactoring everything in the application isn't the purpose of this commit. The purpose is to refactor the file structure and include some get_it implementation. Since this is not our final code (still in experimentation phase), I don't see the point of refactoring everything

app/pubspec.lock
MummenRider: I run the given commands the thing still persist

app/pubspec.lock
MummenRider: Once I deleted the camera in our dependencies, removed the publock and finally run flutter pub get, the min changed to 1.16

app/lib/core/services/service_locator.dart
holtzmak: That seems anti-pattern to me but OK.

app/test/main_test.dart
holtzmak: You should add tests as part of API or #130 . We absolutely require tests on non-experimental code.

holtzmak (approved):
LGTM, your responses to my questions are certainly reasonable. Thank you for continuing to look into best/better practices.

- [x] Please remove `loading_overlay_screen.dart` so we don't accidentally use it.

Issue #131, opened 2021-01-26
Create the History screen


holtzmak:
Blocked by #129 and #127 (I think).

- [x] Create the history screen that will display all the records in the local database
- [x]  (I think an additional function is needed to translate local database records to ATR objects for display, so as to reduce coupling between database and history)

_This screen should not show all of each record's information, determine from the prototypes what information should be previewed._

This screen, at this time, WILL NOT HAVE search functionality for older records, the ability to email mass PDFs, or the ability to select a record and see it's PDF or email it. We'll do additional features later.

HOWEVER, this screen WILL HAVE the ability to select a record and open the history screen for a single record. This work is NOT a part of this task and is the work of #132.

holtzmak:
Added the translation function for ATR object to ATRPreview object in my last PR #160.

There is still a translation function for database ATR object to ATR object, that's part of #130.

PR #144, opened 2021-02-01
Create history screen (no backend connection yet)

holtzmak:
Half of #131 , will need the back end connection yet.

Looks like this! The date has since been formatted after I took this screen shot. The dates are in format: `2020-05-31 05:41` now.

Screen Shot 2021-02-01 at 3 39 08 PM

I've added a preview object for ATRs - we'll be able to reuse this I think with in-progress ATRs if we allow null fields and the be safe with them (for example, maybe we always require a receiver and date so as not to interrupt previews). The `ATWPreViewModel` can also be reused for in-progress items, we'll just filter which items are in the model: in-progress or historical.

The hookup between the view model and the screen using it is so small that it doesn't make sense for us to combine new_travel and history. They can be similar with different buttons for now. Thank you `ListView.builder()`.

I imagine that record selection might be funny with the provider pattern, we specify the `onTap()` for each card to open another widget before we build the widget that holds all the cards (but since business code drives GUI I suppose not a problem).

I also added some test tools in the commit. Our big form is cumbersome to write (easier now with the tools), but I'm struggling with a nicer way to do it: if we split it up more it is hard to change, and if we make one mega-object it becomes anti-pattern to object-oriented code style.


app/lib/core/models/contingency_plan_info.dart
```diff
@@ -42,24 +42,21 @@ class ContingencyPlanInfo {
...
   final List<String> _producerContactsUsed;
   final List<String> _receiverContactsUsed;
   final String disturbancesIdentified;
   final List<ContingencyActivity> _activities;
   final List<String> _actionsTaken;
 
   ContingencyPlanEvent(
-      {@required DateTime eventDate,
-      @required TimeOfDay eventTime,
+      {@required DateTime eventDateAndTime,
```
holtzmak: Combined date and time for many objects, missed it before.

app/lib/ui/widgets/atr_preview.dart
```diff
@@ -0,0 +1,24 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Card(
+      child: Material(
+        child: ListTile(
+          leading: Icon(Icons.local_shipping),
+          title: Text(
+              'Delivery for ${atr.deliveryInfo.recInfo.receiverCompanyName}'),
+          subtitle: Text(
+              'Date: ${DateFormat("yyyy-MM-dd hh:mm").format(atr.vehicleInfo.dateAndTimeLoaded)}, Animals: ${atr.vehicleInfo.animalSpeciesLoaded().join(',')}'),
```
holtzmak: - [x] Now that I look at it... we can just drop `Date: , Animals` because it is obvious from output.

app/test/ui/widgets/atr_preview_test.dart
```diff
@@ -0,0 +1,30 @@
...
+
+void main() {
+  group('Animal Transport Record Preview', () {
+    testWidgets('Previews right information', (WidgetTester tester) async {
+      final testDate = DateTime.now();
+      final testCompany = "Test Company";
+      final testSpecies = "Cows";
+
+      // Must wrap widget in MaterialApp if not "home" widget under test
+      await tester.pumpWidget(new MaterialApp(
```
holtzmak, line 22: Related to [this discussion](https://github.com/flutter/flutter/issues/11859). It's not very obvious, and it doesn't look like they're gonna make it more obvious either.

app/test/ui/widgets/atr_preview_test.dart
```diff
@@ -0,0 +1,30 @@
...
+        atr: testAnimalTransportRecord(
+            deliveryInfo: testDeliveryInfo(
+                recInfo: testReceiverInfo(receiverCompanyName: testCompany)),
+            vehicleInfo: testVehicleInfo(
+                dateAndTimeLoaded: testDate,
+                animalsLoaded: [testAnimalGroup(species: testSpecies)])),
+      )));
+
+      expect(find.text("Delivery for $testCompany"), findsOneWidget);
+      expect(find.text("Date: ${testDate.toString()}, Animals: $testSpecies"),
```
holtzmak: - [x] Oops I broke this test with changing the formatter, will drop Date: and Animals: at the same time

app/test/ui/widgets/atr_preview_test.dart
```diff
@@ -0,0 +1,33 @@
...
+                recInfo: testReceiverInfo(receiverCompanyName: testCompany)),
+            vehicleInfo: testVehicleInfo(
+                dateAndTimeLoaded: testDate,
+                animalsLoaded: [testAnimalGroup(species: testSpecies)])),
+      )));
+
+      expect(find.text("Delivery for $testCompany"), findsOneWidget);
+      expect(
+          find.text(
+              "${DateFormat("yyyy-MM-dd hh:mm").format(testDate)} $testSpecies"),
```
holtzmak, line 28: Usually you don't use production code in the test, but this is a dumb test for a dumb widget.

MummenRider (changes requested):
I know this commit is specifically for History Screen, but since you have a PR open already, 

- [x] please refactor the models as shown below. As you can see, less code here.
```
@immutable
class AcknowledgementInfo {
  final File shipperAck;
  final File transporterAck;
  final File receiverAck;

  AcknowledgementInfo({
    @required this.shipperAck,
    @required this.transporterAck,
    @required this.receiverAck,
  });
```
- [x] I think you refactor the add method in ATRViewModel and make it as Future<void>. Posting should be asynchronous method. We should await if record can be posted or not.
```
 Future<void> add(AnimalTransportRecordPreview animalTransportPreview)  async{
    await _animalTransportPreviews.add(animalTransportPreview);
    notifyListeners();
  }
```

holtzmak:
As per above comment, I will add a placeholder for the database serializer/deserializer/interface service so as to make it obvious that the `add` function is not a `save` function.

holtzmak:
@MummenRider additionally, we will have to keep the initializer list for some fields because of the nature of lists (any fields with `_field` cannot be accessed by `this._field` in the constructor. I should be able to update the other fields.

app/test/ui/widgets/atr_preview_test.dart
```diff
@@ -0,0 +1,32 @@
...
+
+import '../../test_animal_transport_record.dart';
+
+void main() {
+  group('Animal Transport Record Preview', () {
+    testWidgets('Previews right information', (WidgetTester tester) async {
+      final testDate = DateTime.parse("2021-01-01 01:01");
+      final testCompany = "Test Company";
+      final testSpecies = "Cows";
+      final testRecWithDelvInfo = testAnimalTransportRecord().withDeliveryInfo(
```
holtzmak, line 14: Note: We unfortunately cannot use Dart's fantastic cascade operator with the `with()` functions because they return an object.

holtzmak: Also refactored test for easier reading (this is why I made the `with()` functions, better use them >:( )

MummenRider (approved):
LGTM!

PR #145, opened 2021-02-02
Database provider

### LINKED ISSUE #70: Create our first draft of the Database Interface
As per the design in #67 and that we have a specific database type in mind for #69, we should make a generic interface for the different databases that could be used with our software. 

Recall the [Strategy Design Pattern](https://en.wikipedia.org/wiki/Strategy_pattern), we'll want to provide the same pattern in our application so many databases can be used at once with the software. This will help future developers expand our software when picking it up.

Consider the major mobile-interaction databases: SQLite, MariaDB, MongoDB, etc. We'll want an interface that works with the Strategy pattern.

This task is done when:
- [x] An interface is designed that is agnostic for most databases, and referenced in the client-server architecture drawings

SanaKhan1997:
Closes #70.


app/lib/core/models/transporter_info.dart
```diff
@@ -47,4 +47,23 @@ class TransporterInfo {
         trainingExpiryDate = trainingExpiryDate;
 
   List<String> driverNames() => List.unmodifiable(_driverNames);
+
+  Map<String, dynamic> toMap() {
```
holtzmak, line 52: Looks like we'll be reusing this in almost every ATR-related object.

Could we use a library like [this one](https://api.dart.dev/stable/2.8.4/dart-convert/dart-convert-library.html) to make all the `toMap`? 

What worries me about all these functions is that we could miss a field in them, like if you missed `driverNames` below, there's no way you would know immediately until a test or something failed. I read a little on dart forums and it looks like there's no automatic `toMap` for objects yet.

app/lib/core/services/database_provider/db_services.dart
```diff
@@ -0,0 +1,212 @@
...
+import 'package:connectivity/connectivity.dart';
+import 'package:mysql1/mysql1.dart';
+import 'package:app/core/models/user.dart';
+import 'package:app/core/models/animal_transport_record.dart';
+
+class DatabaseServices {
+  var networkStatus;
+  MySqlConnection dbConnect;
+
+  Future<bool> checkConnectivity() async {
```
holtzmak, line 12: I really hope there's either a plugin for this, or Firestore provides this.

app/lib/core/services/database_provider/db_services.dart
```diff
@@ -0,0 +1,212 @@
...
+    var connectivityResult = await (Connectivity().checkConnectivity());
```
holtzmak, line 13: If using `await`, be sure to wrap it in a `try-catch` or some other Future-istic code, especially in the database (we don't do this with tests because they're supposed to work)

app/lib/core/services/database_provider/db_services.dart
```diff
@@ -0,0 +1,212 @@
...
+          newRecord.contingencyInfo.expectedPrepProcess,
+          newRecord.contingencyInfo.goalStatement,
+          newRecord.contingencyInfo.standardAnimalMonitoring,
+          newRecord.contingencyInfo.potentialSafetyActions(),
+          newRecord.contingencyInfo.contingencyEvents(),
+          newRecord.contingencyInfo.crisisContacts(),
+          newRecord.contingencyInfo.potentialHazards()
+        ]);
+
+    await dbConnect.query('''
```
holtzmak, line 112: Each of these should be their own function, it will help us to narrow down the point of failure in the future if something breaks (like we change an object somewhere, or add a field)

app/pubspec.lock
```diff
@@ -648,5 +711,5 @@ packages:
     source: hosted
     version: "2.2.1"
 sdks:
-  dart: ">=2.10.0 <2.11.0"
-  flutter: ">=1.16.0 <2.0.0"
+  dart: ">=2.10.2 <2.11.0"
```
holtzmak, line 714: - [x] Delete `pubspec.lock` and re-run `flutter pub upgrade `to regenerate it.

app/pubspec.yaml
```diff
@@ -34,6 +34,9 @@ dependencies:
   path_provider: ^1.5.1
   pdf: ^2.0.0
   provider: ^4.3.2+3
+  mysql1: ^0.17.1
```
holtzmak, line 37:
```suggestion
  # TODO: #95 or #130. Remove once we've decided on a database (rerun flutter pub get)
  mysql1: ^0.17.1
```

holtzmak (changes requested):
This is a very good first shot at the database! Sorry we keep changing our minds about what database to use (it's good we're doing it now than later at least).

What do you think about making an abstract class for all databases? Like:
```
abstract class DatabaseInterface{
   Future<Record> getRecord(Record rec);
   // and other functions for every query we could make
}
```
That `DatabaseService` or whatever would implement? I think we should extract all your functions and make this now for future developers (or even us if we make changes again).

- [x] Add abstract DatabaseService class that all database services implement


For all my other comments, since this code is experimental and using MySQL, it's likely what you've written will be refactored a lot. Up to you how many changes you want to make.

app/lib/core/services/database_provider/local_db_services.dart
```diff
@@ -0,0 +1,63 @@
...
+class LocalDB {
+  static Database _database;
+  static const String TRANSPORTER = 'TransporterInfo';
+  static const String SHIPPER = 'ShipperInfo';
+  static const String DELIVERY = 'DeliveryInfo';
+  static const String FWR = 'FwrInfo';
+  static const String VEHICLE = 'VehicleLoadingInfo';
+  static const String CONTINGENCY = 'ContingencyPlan';
+  static const String ACKNOWLEDGMENTS = 'AckInfo';
+  static const String DB_NAME = 'humane_transport.db';
```
MummenRider: Why do we nee to define the string here? Can't we create a model for this instead?

MummenRider (commented):
Queries are pretty similar to Firestore, which is good! It looks like you created a bunch of queries, but did you try it to see if it works?

app/lib/core/services/database_provider/local_db_services.dart
holtzmak: Oh yeah, these constants can be in their respective classes

app/lib/core/services/database_provider/db_services.dart
SanaKhan1997: I used connectivity plugin for this so it should work with firestore as well

holtzmak:
@SanaKhan1997 as discussed in backlog refinement, a lot of this code needs to be replaced with Firestore equivalent calls. So, feel free to ignore comments that ask you to refactor code, because this work will be replaced.

Please still do make the abstract class and fix the `pubspec` file.

app/pubspec.lock
holtzmak: Actually, one of the new additions to pubspec.yaml requires a minimum dart and flutter version. This is fine

holtzmak:
Because @SanaKhan1997 rebase got messed up. I opened a [new branch](https://github.com/holtzmak/Humane-Transport/tree/add-abstract-database-interface) with her picked changes. This will either be opened in a new PR, or I will overwrite this branch.

 Sana, let me know that the branch has all your necessary changes please.

PR #146, opened 2021-02-04
Add history integration test

holtzmak:
Continued work for #131 . Did this work to show the test in the vlog.

https://user-images.githubusercontent.com/32527219/106834304-c7328a00-665a-11eb-937d-58b37f91ab9c.mov

I've added the beginnings of the ATR system test "successfully retrieve ATR(s) from remote for specific Transporter". I've added some code to skip login. Not sure how long that will last, however. We should have a test account made up.

Added a little bit of temporary code to `DatabaseService` to make this work, it will be ripped out soon too. This should make it obvious how connections between ViewModels (controllers) and backend works.


app/integration_test/ui/views/history/history_test.dart
```diff
@@ -0,0 +1,34 @@
...
+    await tester.pumpAndSettle();
+  }
+
+  // Some of these are system tests. They will likely be moved from history to system_test dir in the future.
+  group('History screen', () {
+    testWidgets(
+        "successfully retrieve one ATR from remote for specific Transporter",
+        (WidgetTester tester) async {
+      // Ensure app is in good state before testing
+      app.main();
```
holtzmak, line 25: FYI: Running multiple integration tests in one file gets weird because of `getIt` being static for every test. Either we should have each system test in it's own file, or I continue to see if I can't find how to run them all in one.

Doing `setUp() { app.main(); }` does not destroy the previous run instance.

app/lib/core/models/test_animal_transport_record.dart
```diff
@@ -0,0 +1,224 @@
...
+import 'package:app/core/models/delivery_info.dart';
+import 'package:app/core/models/feed_water_rest_info.dart';
+import 'package:app/core/models/loading_vehicle_info.dart';
+import 'package:app/core/models/receiver_info.dart';
+import 'package:app/core/models/shipper_info.dart';
+import 'package:app/core/models/transporter_info.dart';
+import 'package:app/core/utilities/optional.dart';
+import 'package:flutter/material.dart';
+
+// DUPLICATE FILE. TO REMOVE. #130.
```
holtzmak, line 16: This file will absolutely be removed later. I'm just lazy.

app/lib/core/services/database/database_service.dart
```diff
@@ -1,2 +1,14 @@
-// TODO: #130. Add the service which uses the database interface to provide work
-class DatabaseService {}
+import 'dart:async';
+
+import 'package:app/core/models/test_animal_transport_record.dart';
+import 'package:app/ui/widgets/atr_preview.dart';
+
+class DatabaseService {
+  // TODO: #130. Replace with actual DB functions
+  Stream<AnimalTransportRecordPreview> getUpdatedPreviews() async* {
```
holtzmak, line 8: Cheap stream that runs forever. No intensive by any means but this also isn't what we want in actual implementation.

app/test/main_test.dart
```diff
@@ -1,18 +1,24 @@
...
-    setUpAll(() {
-      // TODO: Make a test locator
-      setUpLocator();
-    });
+  group('Application main', () {
+    tearDown(() => resetForTest(testLocator));
 
-    testWidgets('HumaneTransportApp says Welcome', (WidgetTester tester) async {
+    testWidgets('first route is welcome screen', (WidgetTester tester) async {
+      addLazySingletonForTest(testLocator, () => WelcomeViewModel());
```
holtzmak, line 17: Each test should set up only the dependencies it needs. 

Normally, we would mock these, but mocking widgets isn't necessary unless they have logic (which ours don't because we separate that into the ViewModels) and `WelcomeViewModel` has no logic right now (I'm cheap). Similarly with `NavigationService`, it's so simple that the mock would be the same.

However, we should probably inject `AppRouteGenerator` in the future and mock it.
 
 _This is more of an integration test, using both `HumaneTransportApp()` and `Welcome`, if we mocked `ARG` then all we would be testing is that `HTA` calls on it, which is the promise of `MaterialApp(onGenerateRoute: ...` and makes for a pretty dumb test._

MummenRider (approved):
LGTM!

PR #148, opened 2021-02-05
Finished firebase api exp

### LINKED ISSUE #95: Experiment with the Firebase API
Follow-up to #70 but with Firebase. Blocked by #96 (invalid if the task chooses SQlite/AWS)

You should investigate first the work done so far in #70, and see what changes should be made in order to use Firebase. Additionally, some investigation done in #81 could be useful. Speak with the assignees of the two tasks before beginning.

This task is done when:
- [x] You can use/understand how the Firebase API lets us make fetches, updates, and any other notable things you find

MummenRider:
Upgraded Flutter to 1.22.6 [Stable](https://flutter.dev/docs/development/tools/sdk/releases) version

Here's what I did to upgrade:
- ` flutter channel stable`  (I was in dev). This will also show if you're behind from Flutter latest stable version
- `flutter upgrade`
- `flutter clean`
- `flutter run`

Note: 
- clark@gmail.com will not work anymore. Accounts to log in:
mumenrider@gmail.com and  mansipatel@me.ca, pass for both account is  123456.
- Circular indicator during fresh start is needed for testing. It also make sense (for me at least) to put some loading indicator there that the page is working. Rather than just a static image.
- Alert dialogs is set up  similar to navigation service. Followed the tutorial from [FilledStack](https://www.filledstacks.com/post/show-dialogs-without-context/)

https://user-images.githubusercontent.com/43891568/107104539-c3803e00-67e7-11eb-8d1c-1b17e5f4e370.mov

### DIFF
--- a/.github/workflows/flutter-action-build-and-integration-testing.yml
+++ b/.github/workflows/flutter-action-build-and-integration-testing.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Set up Flutter actions
         uses: subosito/flutter-action@v1
         with:
-          flutter-version: '1.22.5'
+          flutter-version: '1.22.6'
           channel: 'stable'
       - name: List all available devices
         run: xcrun instruments -s
--- a/app/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
+++ b/app/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:Runner.xcodeproj">
+      location = "self:">
    </FileRef>
 </Workspace>
--- a/app/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
+++ b/app/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "self:">
+      location = "group:Runner.xcodeproj">
    </FileRef>
 </Workspace>
--- a/app/lib/core/enums/auth_result_status.dart
+++ b/app/lib/core/enums/auth_result_status.dart
@@ -1,0 +1,10 @@
+enum ResultStatus {
+  successful,
+  isLoggedIn,
+  emailAlreadyUsed,
+  wrongPassword,
+  invalidEmail,
+  userNotFound,
+  weakpassword,
+  undefined,
+}
--- a/app/lib/core/models/dialogs.dart
+++ b/app/lib/core/models/dialogs.dart
@@ -1,0 +1,26 @@
+import 'package:flutter/foundation.dart';
+
+class DialogRequest {
+  final String title;
+  final String description;
+  final String buttonText;
+  final String cancelText;
+
+  DialogRequest(
+      {@required this.title,
+      @required this.description,
+      @required this.buttonText,
+      this.cancelText});
+}
+
+class DialogResponse {
+  final String fieldOne;
+  final String fieldTwo;
+  final bool confirmed;
+
+  DialogResponse({
+    this.fieldOne,
+    this.fieldTwo,
+    this.confirmed,
+  });
+}
--- a/app/lib/core/models/test_history.dart
+++ b/app/lib/core/models/test_history.dart
@@ -1,0 +1,30 @@
+import 'package:flutter/material.dart';
+
+class HistoryRecord {
+  final String title;
+  final String userId;
+  final String documentId;
+  HistoryRecord({
+    @required this.title,
+    @required this.userId,
+    this.documentId,
+  });
+
+  Map<String, dynamic> toJson() {
+    return {
+      'userId': userId,
+      'title': title,
+    };
+  }
+
+  static HistoryRecord fromFirestore(
+      Map<String, dynamic> data, String documentId) {
+    if (data == null) return null;
+
+    return HistoryRecord(
+      title: data['title'],
+      userId: data['userId'],
+      documentId: documentId,
+    );
+  }
+}
--- a/app/lib/core/models/test_history.dart
+++ b/app/lib/core/models/test_history.dart
@@ -10,21 +10,15 @@ class HistoryRecord {
     this.documentId,
   });
 
+  HistoryRecord.fromFirestore(Map<String, dynamic> data, String documentId)
+      : title = data['title'],
+        userId = data['userId'],
+        documentId = documentId;
+
   Map<String, dynamic> toJson() {
     return {
       'userId': userId,
       'title': title,
     };
-  }
-
-  static HistoryRecord fromFirestore(
-      Map<String, dynamic> data, String documentId) {
-    if (data == null) return null;
-
-    return HistoryRecord(
-      title: data['title'],
-      userId: data['userId'],
-      documentId: documentId,
-    );
   }
 }
--- a/app/lib/core/models/user_info.dart
+++ b/app/lib/core/models/user_info.dart
@@ -1,0 +1,34 @@
+import 'package:flutter/material.dart';
+
+class UserInformation {
+  final String userId;
+  final String firstName;
+  final String lastName;
+  final String userEmailAddress;
+  final String userPhoneNumber;
+
+  UserInformation({
+    @required this.userId,
+    @required this.firstName,
+    @required this.lastName,
+    @required this.userEmailAddress,
+    @required this.userPhoneNumber,
+  });
+
+  UserInformation.fromFirestore(Map<String, dynamic> data)
+      : userId = data['userId'],
+        firstName = data['firstName'],
+        lastName = data['lastName'],
+        userEmailAddress = data['userEmailAddress'],
+        userPhoneNumber = data['userPhoneNumber'];
+
+  Map<String, dynamic> toJson() {
+    return {
+      'userId': userId,
+      'firstName': firstName,
+      'lastName': lastName,
+      'userEmailAddress': userEmailAddress,
+      'userPhoneNumber': userPhoneNumber,
+    };
+  }
+}
--- a/app/lib/core/services/authentication/auth_service.dart
+++ b/app/lib/core/services/authentication/auth_service.dart
@@ -1,38 +1,108 @@
 import 'dart:async';
+import 'package:app/core/models/user_info.dart';
+import 'package:app/core/services/firestore/firestore_service.dart';
+import 'package:app/core/services/service_locator.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
 
 class AuthenticationService {
   final FirebaseAuth _firebaseAuth = FirebaseAuth.instance;
+  final FirestoreService _firestoreService = locator<FirestoreService>();
 
+  UserInformation _currentUser;
+  UserInformation get currentUser => _currentUser;
+
+  /*
+   This needs to be a dynamic type as we may expect bool or a string as return type.
+   Note: Same logic applies to other methods here.
+  [FirebaseAuthException] - Will return String if something went wrong
+  */
   Future signIn({
     @required String email,
     @required String password,
   }) async {
     try {
-      var user = await _firebaseAuth.signInWithEmailAndPassword(
+      UserCredential authResult =
+          await _firebaseAuth.signInWithEmailAndPassword(
         email: email,
         password: password,
       );
-      return user != null;
-    } catch (e) {
-      return e.message;
+      var user = await _firestoreService.fetchUser(authResult.user.uid);
+      var isUserExist = user != null;
+
+      await _populateCurrentUser(authResult.user);
+
+      print('${user.firstName} currently signed in!');
+      return isUserExist;
+    } on FirebaseAuthException catch (e) {
+      return e.code;
     }
   }
 
-  Future signOut() async {
-    // TODO: Should handle exception error
-    await _firebaseAuth.signOut();
+  Future signUp({
+    @required String firstName,
+    @required String lastName,
+    @required String userEmailAddress,
+    @required String userPhoneNumber,
+    @required String password,
+  }) async {
+    try {
+      UserCredential authResult =
+          await _firebaseAuth.createUserWithEmailAndPassword(
+        email: userEmailAddress,
+        password: password,
+      );
+
+      // Uses the service to create a new user to Firestore
+      await _firestoreService.addUserToFirestore(UserInformation(
+        userId: authResult.user.uid,
+        firstName: firstName,
+        lastName: lastName,
+        userEmailAddress: userEmailAddress,
+        userPhoneNumber: userPhoneNumber,
+      ));
+
+      return authResult.user != null;
+    } on FirebaseAuthException catch (e) {
+      return e.code;
+    }
   }
 
-  void user() {
-    // ignore: deprecated_member_use
-    FirebaseAuth.instance.authStateChanges().listen((User user) {
-      if (user == null) {
-        return null;
-      } else {
-        return user;
+  Future<void> signOut() async {
+    try {
+      await _firebaseAuth.signOut();
+    } on FirebaseAuthException catch (e) {
+      return e.code;
+    }
+  }
+
+  Future isLoggedIn() async {
+    var firebaseUser = _firebaseAuth.currentUser;
+    if (firebaseUser == null) return false;
+
+    try {
+      var user = await _firestoreService.fetchUser(firebaseUser.uid);
+      var isUserExist = user != null;
+
+      /*
+      As soon as the system confirmed a user is logged in, populate
+      user information to be readily available.
+      */
+      await _populateCurrentUser(firebaseUser);
+
+      return isUserExist;
+    } on FirebaseAuthException catch (e) {
+      return e.code;
+    }
+  }
+
+  Future _populateCurrentUser(User user) async {
+    try {
+      if (user != null) {
+        _currentUser = await _firestoreService.fetchUser(user.uid);
       }
-    });
+    } on FirebaseAuthException catch (e) {
+      return e.code;
+    }
   }
 }
--- a/app/lib/core/services/authentication/auth_service.dart
+++ b/app/lib/core/services/authentication/auth_service.dart
@@ -1,23 +1,20 @@
 import 'dart:async';
+import 'package:app/core/enums/auth_result_status.dart';
 import 'package:app/core/models/user_info.dart';
 import 'package:app/core/services/firestore/firestore_service.dart';
 import 'package:app/core/services/service_locator.dart';
+import 'package:app/core/utilities/auth_exception.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
 
 class AuthenticationService {
   final FirebaseAuth _firebaseAuth = FirebaseAuth.instance;
   final FirestoreService _firestoreService = locator<FirestoreService>();
-
+  ResultStatus _status;
   UserInformation _currentUser;
   UserInformation get currentUser => _currentUser;
 
-  /*
-   This needs to be a dynamic type as we may expect bool or a string as return type.
-   Note: Same logic applies to other methods here.
-  [FirebaseAuthException] - Will return String if something went wrong
-  */
-  Future signIn({
+  Future<ResultStatus> signIn({
     @required String email,
     @required String password,
   }) async {
@@ -28,18 +25,20 @@ class AuthenticationService {
         password: password,
       );
       var user = await _firestoreService.fetchUser(authResult.user.uid);
-      var isUserExist = user != null;
 
-      await _populateCurrentUser(authResult.user);
-
-      print('${user.firstName} currently signed in!');
-      return isUserExist;
+      if (user != null) {
+        await _populateCurrentUser(authResult.user);
+        _status = ResultStatus.successful;
+      } else {
+        _status = ResultStatus.undefined;
+      }
     } on FirebaseAuthException catch (e) {
-      return e.code;
+      _status = ExceptionHandler.handleException(e);
     }
+    return _status;
   }
 
-  Future signUp({
+  Future<ResultStatus> signUp({
     @required String firstName,
     @required String lastName,
     @required String userEmailAddress,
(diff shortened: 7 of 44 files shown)


app/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
```diff
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:Runner.xcodeproj">
+      location = "self:">
```
holtzmak: This seems suspicious.

app/lib/core/models/test_history.dart
```diff
@@ -0,0 +1,30 @@
...
+  Map<String, dynamic> toJson() {
+    return {
+      'userId': userId,
+      'title': title,
+    };
+  }
+
+  static HistoryRecord fromFirestore(
+      Map<String, dynamic> data, String documentId) {
+    if (data == null) return null;
```
holtzmak: No nulls please. We should return an `Optional` if it could be null.

- [x] Make return type `Optional<HistoryRecord>`.

app/lib/core/models/user_info.dart
```diff
@@ -0,0 +1,34 @@
...
+
+  UserInformation({
+    @required this.userId,
+    @required this.firstName,
+    @required this.lastName,
+    @required this.userEmailAddress,
+    @required this.userPhoneNumber,
+  });
+
+  UserInformation.fromFirestore(Map<String, dynamic> data)
```
holtzmak, line 18: We check if data is null in `HistoryRecord`. Should we not also be doing that here?

app/lib/core/models/test_history.dart
holtzmak: Also, this should not be a `static` method. Let's add this as a named constructor `HistoryRecord.fromFirestore`.

app/lib/core/services/authentication/auth_service.dart
```diff
@@ -1,38 +1,108 @@
...
+      // Uses the service to create a new user to Firestore
+      await _firestoreService.addUserToFirestore(UserInformation(
+        userId: authResult.user.uid,
+        firstName: firstName,
+        lastName: lastName,
+        userEmailAddress: userEmailAddress,
+        userPhoneNumber: userPhoneNumber,
+      ));
+
+      return authResult.user != null;
```
holtzmak: Instead of returning all these different things, let's make a controlled return type. If you've heard of `Try` and `Failure`, this should be familiar. How about:

- [x] Add the following and use them

```
enum ResponseType { success, failure, catastrophic_failure }

class Response<T> { 
    final ResponseType type;
    final String reason;
    final Optional<T> response;

    Response.failure({@required this.type, @required this.reason}) :
    response = Optional.empty();

    Response.success({@required this.response}):
    type = ResponseType.success,
    reason = "Successful";
}
```
This will help us to have very clean and thorough error handling for future work.

app/lib/core/services/authentication/auth_service.dart
```diff
@@ -1,38 +1,108 @@
...
+      return e.code;
     }
   }
 
-  Future signOut() async {
-    // TODO: Should handle exception error
-    await _firebaseAuth.signOut();
+  Future isLoggedIn() async {
+    var firebaseUser = _firebaseAuth.currentUser;
+    if (firebaseUser == null) return false;
```
holtzmak, line 85: I hate null. 
- [x] Let's add (or Kelly to add) `Optional.fromNullable` to our `Optional` class

app/lib/core/services/firestore/firestore_service.dart
```diff
@@ -0,0 +1,87 @@
+import 'dart:async';
+import 'package:app/core/models/test_history.dart';
+import 'package:app/core/models/user_info.dart';
+import 'package:cloud_firestore/cloud_firestore.dart';
+
+class FirestoreService {
+  final CollectionReference _userCollectionReference =
+      FirebaseFirestore.instance.collection('users');
```
holtzmak, line 10: Is there a way to programmatically state these collections? I'm wondering if we're sharing something secret here.

app/lib/core/services/navigation/nav_service.dart
```diff
@@ -9,7 +9,12 @@ class NavigationService {
...
   }
 
-  Future<dynamic> navigateTo(String routeName) {
-    return _navigationKey.currentState.pushNamed(routeName);
+  Future<dynamic> navigateTo(String routeName, {dynamic arguments}) {
+    return _navigationKey.currentState
+        .pushNamed(routeName, arguments: arguments);
+  }
+
+  Future<dynamic> navigateAndReplace(String routeName) {
```
holtzmak, line 17: Would this be useful for the PDF and email situation we have in `pdf_preview.dart`?

app/lib/core/services/service_locator.dart
```diff
@@ -1,26 +1,35 @@
...
 import 'package:app/core/view_models/welcome_view_model.dart';
 import 'package:get_it/get_it.dart';
-
 import 'authentication/auth_service.dart';
+import 'firestore/firestore_service.dart';
 import 'navigation/nav_service.dart';
 
 final locator = GetIt.instance;
 
 void setUpLocator() {
```
holtzmak, line 20: I'm wondering is we should give this a better name and split the singletons from the factories (and add comments to tell people when to use what).

app/lib/core/view_models/add_history_view_model.dart
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+      result = await _firestoreService.updatHistoryTravel(HistoryRecord(
+          title: title,
+          userId: _edittingHistory.userId,
+          documentId: _edittingHistory.documentId));
+    }
+
+    setBusy(false);
+
+    // If result is type String, then something went wrong
```
holtzmak, line 36: Our new error types you will add ( :) ) should make this better!

app/lib/core/view_models/base_view_model.dart
```diff
@@ -1,11 +1,26 @@
...
-  ViewState _state = ViewState.Idle;
-  ViewState get state => _state;
-  void setState(ViewState viewState) {
-    _state = viewState;
+  final AuthenticationService _authenticationService =
+      locator<AuthenticationService>();
+
+  final NavigationService _navigationService = locator<NavigationService>();
+
+  UserInformation get currentUser => _authenticationService.currentUser;
```
holtzmak, line 13: Does every change notifier need the current user? This smells.

app/lib/core/view_models/signup_view_model.dart
```diff
@@ -0,0 +1,46 @@
...
+      if (result) {
+        popScreen();
+      } else {
+        //Otherwise, we show the error message
+        print('Please try again later');
+      }
+    } else {
+      //If it's not a bool type, then something went wrong
+      //Checks if password is correct, email is formatted and etc
+      print(result);
```
holtzmak: All these prints should be bubble up into alert dialogs instead of just the build output.

- [x] What do you think about replacing the prints with `alertDialogs`? It should be more obvious what's wrong for the user

app/lib/core/view_models/welcome_view_model.dart
```diff
@@ -1,3 +1,18 @@
+import 'package:app/core/services/navigation/nav_service.dart';
+import 'package:app/core/services/service_locator.dart';
 import 'package:app/core/view_models/base_view_model.dart';
+import 'package:app/ui/views/signin/sign_in.dart';
+import 'package:app/ui/views/signup/signup.dart';
 
-class WelcomeViewModel extends BaseViewModel {}
+// TODO: To Mansi - Add something if needed
```
holtzmak, line 7: - [x] Add a task number for this. What is to be added?

app/test/ui/widgets/firebase_auth_mock.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
```
holtzmak, line 2: Can we have BSD license with our GPL? We need to watch what code we use in our project. There's probably a tool that will get all the licenses for our plugins but we'll need to manually the license for code we use like this.

- [x] Verify we can use this code and track the license somewhere

For example, we'll have an `about` button in the side drawer that will list all the licenses we used in our software. For further example, see our university webmail. There is an `about` button in the bottom-left corner that shows what we should do if we have multiple licenses.

holtzmak (changes requested):
Lot's of requested changes here (they're relatively minor as I don't have concerns about the UI or business code separation, mostly error handling improvements I want to see).

app/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
MummenRider: Ill try to change this back to group and see what happens.  I did not touch this file whatsoever, maybe during flutter upgrade

app/lib/core/models/test_history.dart
MummenRider: I was just testing another way to read data from firestore, I intended to change this but I completely forgot about it. I will change this in my next commit

app/lib/core/models/user_info.dart
MummenRider: This applies to services already and therefore not need to be in the model itself. We check if snapshot has data, if it's not null, then we call fromFirestore

holtzmak: I'm sorry, could you explain further? I'm not sure what "this applies to services already" means.

holtzmak:
- [x] Please also update the `github/worklfows/` YML files with the new flutter version.

For example, in `flutter-action-...,yml`:
```
      - name: Set up Flutter actions
        uses: subosito/flutter-action@v1
        with:
          flutter-version: '1.22.5'  # Should now be 1.22.6?
          channel: 'stable'
```

app/lib/core/services/authentication/auth_service.dart
MummenRider: Will implement a better error handling in my next commit

app/lib/core/view_models/add_history_view_model.dart
MummenRider: Actually, we still need to know the current state of user if it's awaiting or not. I'm not sure how I can make this any better nor I see connection with the error messages.

app/lib/core/view_models/base_view_model.dart
MummenRider: You need to know if there is a current user in most of viewmodels that communicates to firestore services. I put it in the base model, because we will have multiple screen that  needs to know the current user. A great example would be the AddHistoryViewModel

app/lib/core/view_models/signup_view_model.dart
MummenRider: I will create a dialog service for this in my next commit

app/lib/core/view_models/welcome_view_model.dart
MummenRider: This is not defined task right now, but @mansiPaTeLL is creating a logo that will be put in welcomeview screen. To do that, I assume we might need some kind of functionality inside view model

app/lib/core/models/user_info.dart
MummenRider: I'm referring to services like authentication when it's checking if userinfo/ Firebase user is null or not.

holtzmak: I understand what you're saying as: "we're already checking if it's null in another class". I think we should be consistent with null checks: either we check if null before making the class, or do it in the constructor. 

- [x] So, we should probably not accept null data in `HistoryRecord fromFirestore()` and check that the data is not null before using it.

app/lib/core/services/authentication/auth_service.dart
holtzmak: I will add this another time.

app/lib/core/view_models/add_history_view_model.dart
holtzmak: Well, you're checking if the result is a `String` type, which is a smelly way of checking if something went wrong. You should either:
- [x] Check the contents of the string for a notable reason that the result is a failure, or
- [ ] Use the `Response` class I suggested to make this more understandable:

```
var Response response;
// set response
response.isFailure() ? print('Something went wrong: $response.message') : print('Successfully added');

Comment (holtzmak) [reply app/lib/core/view_models/add_history_view_model.dart:36]: Or even better:
```
response.ifFailure((res) => print('Had a problem adding history: ${res.error}')
                .ifSuccess((res) => print('Successfully added history');

app/lib/core/view_models/base_view_model.dart
holtzmak: So... that seems anti-pattern to me. Some classes don't need this information, so we shouldn't have it "globally" available to all extensions.

We use the dependency injection service so we don't have to have dependencies in the base class like this.

- [x] Make it so only the classes that need the service have the service

app/lib/core/view_models/welcome_view_model.dart
holtzmak: OK. I want to be more picky about how many TODOs we have so we don't accidentally miss something.

.github/workflows/flutter-action-build-and-integration-testing.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Set up Flutter actions
         uses: subosito/flutter-action@v1
         with:
-          flutter-version: '1.22.5'
+          flutter-version: '1.22.6'
```
holtzmak, line 36: - [x] We have two workflow files, please also change `flutter-action-format-and-unit-testing.yml`

app/lib/core/enums/auth_result_status.dart
```diff
@@ -0,0 +1,10 @@
+enum ResultStatus {
```
holtzmak, line 1: _We had a discussion over discord about this, I'll make a note here as a reminder:_

It's likely this enum (and the other file with all the possible messages) won't be here to stay - I don't think it's scalable, and since we don't define failure response codes for external systems I think it simple enough to write the failure messages where they happen, and the two types of response (success, failure) is enough for us:

Example of how I intended `Response` to be used:
```
// some function that could fail
return Future.wait([futureA, futureB, futureC, futureD])
                      .then((a, b, c, d) => // do work with a,b,c,d... then return Response.success(newThing)
                      .onError((err) => // handle error, return Response.failure(err.message));
```
The `Response` is a useful wrapper between `Future<T>` and the type `T` it holds

app/lib/core/models/dialogs.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:flutter/foundation.dart';
+
+class DialogRequest {
```
holtzmak, line 3: Could we not have used the `alert_dialog.dart` that I added a while back?

app/lib/core/services/authentication/auth_service.dart
```diff
@@ -1,23 +1,20 @@
...
 import 'package:app/core/services/service_locator.dart';
+import 'package:app/core/utilities/auth_exception.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
 
 class AuthenticationService {
   final FirebaseAuth _firebaseAuth = FirebaseAuth.instance;
   final FirestoreService _firestoreService = locator<FirestoreService>();
-
+  ResultStatus _status;
```
holtzmak, line 13: Why keep the last result code here? I wouldn't think we need to check this, unless we kill the service upon any failure. The functions returns are enough.

app/lib/core/services/authentication/auth_service.dart
```diff
@@ -53,56 +52,56 @@ class AuthenticationService {
...
       await _firebaseAuth.signOut();
     } on FirebaseAuthException catch (e) {
-      return e.code;
+      _status = ExceptionHandler.handleException(e);
     }
+    return _status;
   }
 
-  Future isLoggedIn() async {
+  Future<dynamic> isLoggedIn() async {
```
holtzmak, line 83: - [x] Probably should just return true or false here

app/lib/core/view_models/add_history_view_model.dart
MummenRider: The code highlighted to your review is `setBusy(false)`. So, I thought you were referring to that code. In my recent change, I did not check if the result is in type String. Instead, I checked if the response from ResultStatus is successful or now. 
```
 if (result == ResultStatus.successful) {
      await _dialogService.showDialog(
        title: 'Travel History',
        description: 'Successfully added!',
      );
    } else {
      final errorMessage = ExceptionHandler.exceptionMessage(result);
      await _dialogService.showDialog(
        title: 'Travel History Creation Failed',
        description: errorMessage,
      );
    }
```
I can simply use a shorthand if-else  here, but I honestly think this is readable and understandable this way

app/lib/humane_transport_app.dart
```diff
@@ -17,6 +19,11 @@ class _HumaneTransportAppState extends State<HumaneTransportApp> {
       initialRoute: SplashScreen.route,
       onGenerateRoute: AppRouteGenerator.onGenerateRoute,
       navigatorKey: locator<NavigationService>().navigationKey,
+      builder: (context, child) => Navigator(
```
holtzmak, line 22: The navigator is a dialog service???

holtzmak (changes requested):
Some confusion with your changes, I've asked questions. There are also questions from my last review I'd like to read your answers for. Be wary to expand this and see the questions there:
Screen Shot 2021-02-05 at 8 08 36 PM

app/lib/core/view_models/add_history_view_model.dart
holtzmak: Oh, the code for the review comment is the line immediately before the comment:
Screen Shot 2021-02-05 at 8 13 59 PM

app/lib/core/services/firestore/firestore_service.dart
MummenRider: I don't think so... we need to specially state which collection are you referring to. If you find other way how to reference a collection, please let me know.

app/lib/core/services/navigation/nav_service.dart
MummenRider: Yes, I added this just in case we need to do navigation like this.

app/lib/core/services/service_locator.dart
MummenRider: We can definitely change the naming in the future

app/lib/core/view_models/base_view_model.dart
MummenRider: That is indeed a really good point! I'll change this in my next commit.

.github/workflows/flutter-action-build-and-integration-testing.yml
MummenRider: Will do, thanks!

app/lib/core/models/dialogs.dart
MummenRider:
```
launchAlertDialog(BuildContext context,
        {String title = "Title", String warning = "Warning"}) =>
    showDialog(
        context: context,
        builder: (context) => AlertDialog(
              title: Text(title),
              content: Text(warning),
            ));
```
As you can see, this accepts a context. But since we want a total separation of UI to business logic, we lose the ability to get the context. Hence to why we use Globalkey instead, which is can be used a sub. for context.

app/lib/core/services/authentication/auth_service.dart
MummenRider: The code being highlighted is this `ResultStatus _status;`, so I think you're referring to this one. `_status ` holds the value whether the it's a success or a failure. 
 ```
var result = await _authenticationService.signIn(
      email: email,
      password: password,
    );
```
The code above is used in viewmodel, and that result will contain whether it's a success or not.

File: app/lib/humane_transport_app.dart (hunk 19 above)
Comment (MummenRider) [reply app/lib/humane_transport_app.dart:22]: FilledStack - _It's common that the BuildContext is not available where you business logic is, whether you're using BLoC, pure provider, redux, get_it only, get, etc. It's easier in terms of writing code to be able to show your dialogs from where your failures / success are happening, which is why the DialogService was born._

Comment (holtzmak) [reply app/lib/humane_transport_app.dart:22]: Can you send me the link you got this from? I can guess that the material app builder of widgets uses dialog manager somehow, but I don't understand how.

File: app/lib/core/services/authentication/auth_service.dart (hunk 17 above)
Comment (holtzmak) [reply app/lib/core/services/authentication/auth_service.dart:13]: What I don't understand about this, is that your functions already return the status and then that return is used. I don't see the need to have it as a global class member/field?

File: app/lib/core/enums/auth_result_status.dart (hunk 15 above)
Comment (MummenRider) [reply app/lib/core/enums/auth_result_status.dart:1]: @holtzmak ... though the example you gave me an idea how to use then catch, instead of try catch, I still don't know how I can incorporate the "Optional" my current code. 

```
enum ResponseType { success, failure, catastrophic_failure }

class Response<T> { 
    final ResponseType type;
    final String reason;
    final Optional<T> response;

    Response.failure({@required this.type, @required this.reason}) :
    response = Optional.empty();

    Response.success({@required this.response}):
    type = ResponseType.success,
    reason = "Successful";
}
```

The exception is simple enough for me to understand, but how about the response type? Using what I have right now, can you give an example how to do this? Let's use the signIn method as an example

 ```
Future<ResultStatus> signIn({
    @required String email,
    @required String password,
  }) async {
    try {
      UserCredential authResult =
          await _firebaseAuth.signInWithEmailAndPassword(
        email: email,
        password: password,
      );
      var user = await _firestoreService.fetchUser(authResult.user.uid);

      if (user != null) {
        await _populateCurrentUser(authResult.user);
        _status = ResultStatus.successful;
      } else {
        _status = ResultStatus.undefined;
      }
    } on FirebaseAuthException catch (e) {
      _status = ExceptionHandler.handleException(e);
    }
    return _status;
  }
```

Then, the viewmodel is this:
```
 Future signIn({
    @required String email,
    @required String password,
  }) async {
    setBusy(true);
    var result = await _authenticationService.signIn(
      email: email,
      password: password,
    );
    setBusy(false);
    if (result == ResultStatus.successful) {
      _navigationService.navigateTo(HomeRootScreen.route);
    } else {
      final errorMessage = ExceptionHandler.exceptionMessage(result);
      await _dialogService.showDialog(
        title: 'Sign up failed',
        description: errorMessage,
      );
    }
  }
```

My attempt:
```
 Future signIn({
    @required String email,
    @required String password,
  }) async {
    _firebaseAuth
        .signInWithEmailAndPassword(email: email, password: password)
        .then((authResult) {
      _populateCurrentUser(authResult.user);
      /*
      How to use the Response.success? I want to return the authResult.user
      if it's not equal to null
      */
      return authResult.user != null;
    }).catchError((e) =>
            Response.failure(type: ResponseType.failure, reason: e.code));
  }
```

My viewmodel:

```
  Future signIn({
    @required String email,
    @required String password,
  }) async {
    setBusy(true);
    _authenticationService
        .signIn(email: email, password: password)
        .then((result) => _navigationService.navigateTo(HomeRootScreen.route))
        .catchError((error) => _dialogService.showDialog(
              title: 'Sign up failed',
              description: error,
            ));
    setBusy(false);
  }
```

app/lib/core/services/authentication/auth_service.dart
MummenRider: I made it global is don't need keep calling `ResultStatus` every time I need to assign a value (like success or not). ` _status = ResultStatus.successful;` I still need to do this to get the values from enums

app/lib/humane_transport_app.dart
MummenRider: I provided the link in my commit notes, but here it  is: [FilledStack](https://www.filledstacks.com/post/show-dialogs-without-context/)

app/lib/core/enums/auth_result_status.dart
holtzmak: I think your view model is right, as long as the dialog appears over the busy overlay. 

In your attempt, I expect the return you want is something like this:
```
.then((authResult) {
final result = Optional(authResult.user);
// Probably should hold optional current user rather than null, too
_populateCurrentUser(result);
// The statement `authResult.user != null returns a bool, right
// Otherwise, if you're using the user, do Response.success(result)
return result.isPresent() ? Response.success(true) 
    : Response.failure(type: ResponseType.failure, reason: "No such account exists");
}
})

File: app/lib/core/models/dialogs.dart (hunk 16 above)
Comment (holtzmak) [reply app/lib/core/models/dialogs.dart:3]: Should we remove `launchAlertDialog` now, to prevent people from using it instead of `DialogService`?

File: app/lib/core/services/authentication/auth_service.dart (hunk 17 above)
Comment (holtzmak) [reply app/lib/core/services/authentication/auth_service.dart:13]: Hmm, OK. I am still confused about this - why not return the result status from the function, the status would tell the using code if the work was successful or not.

File: app/lib/humane_transport_app.dart (hunk 19 above)
Comment (holtzmak) [reply app/lib/humane_transport_app.dart:22]: In the example in your link:
```
class MyApp extends StatelessWidget {
  @override
    return MaterialApp(
      title: 'Flutter Demo',
      home: DialogExampleView(),
      onGenerateRoute: Router().onGenerateRoute,
      // Set the navigatorKey
      navigatorKey: locator<DialogService>().navigatorKey,
    );
  }
}
```

It looks like the navigator key is the dialog service's. We have separate navigator and dialog service.

What I don't understand is that you've wrapped our dialog manager in a navigator. Why?

**Comment by holtzmak** [comment]
@MummenRider I can help resolve conflicts if not easy.

File: app/lib/humane_transport_app.dart (hunk 19 above)
Comment (MummenRider) [reply app/lib/humane_transport_app.dart:22]: Check this [article](https://medium.com/flutter-community/manager-your-flutter-dialogs-with-a-dialog-manager-1e862529523a) and it explains everything about dialog service and why he chose to wrap it with Navigator.

Comment (holtzmak) [reply app/lib/humane_transport_app.dart:22]: Ah, OK. Makes sense now. Could you comment that article in this file?

Comment (MummenRider) [reply app/lib/humane_transport_app.dart:22]: will do!

File: app/lib/core/models/dialogs.dart (hunk 16 above)
Comment (MummenRider) [reply app/lib/core/models/dialogs.dart:3]: Will do in my next commit

File: app/lib/core/enums/auth_result_status.dart (hunk 15 above)
Comment (MummenRider) [reply app/lib/core/enums/auth_result_status.dart:1]: For now, is it okay to push my current responses? 
`Response.success(true) `
The code above is giving me an error. I also run into this problem before, it looks like response success isn't accepting bool. So, we can't just assign true value to it

Error Message:
```
{Optional<dynamic> response}
The argument type 'bool' can't be assigned to the parameter type 'Optional<dynamic>'.dart(argument_type_not_assignable)
```

MummenRider:
I will try to solve it on my own and if I require assistance I will let you know. Thanks!
> @MummenRider I can help resolve conflicts if not easy.

PR #149, opened 2021-02-06
Add single history screen

### LINKED ISSUE #132: Create the single ATR History screen
Blocked by #129, #127, and #131.

- [x] Create the history screen that will display one of the selected records from the main history screen

_This screen should show all of the record's information._

This screen, at this time, WILL NOT HAVE the ability see it's PDF or email it. We'll do additional features later.

holtzmak:
Closes #132 .

Preview of the changes:

https://user-images.githubusercontent.com/32527219/107104748-9c763c00-67e8-11eb-869c-43042383142e.mov

One notable thing: the text doesn't look good. When we go to clean up the GUI appearance we should render the text nicer.


MummenRider (approved):
LGTM!

app/lib/core/models/acknowledgement_info.dart
```diff
@@ -12,4 +12,11 @@ class AcknowledgementInfo {
       {@required this.shipperAck,
       @required this.transporterAck,
       @required this.receiverAck});
+
+  // TODO: Resolve how to display ack info
+  // Most other info is strings, perhaps this should give a widget for display
+  String toString() => '''Shipper acknowledgement:
```
holtzmak, line 18: The more I look at these methods the more I hate them. We will absolutely be extracting these all out to their respective display widget (which increases the amount of code to maintain, but the display and object are 1-to-1 so changing one forces changes on the other) (I'm still not sure of a better way to do this...).

[Expandable list of widgets example.](https://stackoverflow.com/questions/50530152/how-to-create-expandable-listview-in-flutter)

holtzmak: @MummenRider should I do this right away, or move on?

PR #157, opened 2021-02-13
Dialog service 

MummenRider:
All credit goes to [FilledStacks](https://www.filledstacks.com/post/manager-your-flutter-dialogs-with-a-dialog-manager/) by creating dialog service that can be used in business logic side.


app/lib/core/services/dialog/dialog_service.dart
```diff
@@ -0,0 +1,50 @@
...
+        description: description,
+        buttonText: confirmationText,
+        cancelText: cancelText));
+    return _dialogCompleter.future;
+  }
+
+  void dialogComplete(DialogResponse response) {
+    _dialogNavigationKey.currentState.pop();
+    _dialogCompleter.complete(response);
+    _dialogCompleter = null;
```
holtzmak: Needs to be null? Can we use `Optional`?

app/lib/core/services/dialog/dialog_service.dart
```diff
@@ -0,0 +1,50 @@
+// Followed tutorial: https://www.filledstacks.com/post/manager-your-flutter-dialogs-with-a-dialog-manager/
+import 'dart:async';
+import 'package:app/core/models/dialog.dart';
+import 'package:flutter/material.dart';
+
+class DialogService {
+  GlobalKey<NavigatorState> _dialogNavigationKey = GlobalKey<NavigatorState>();
```
holtzmak: Can we make this final?

app/lib/core/utilities/dialog_manager.dart
```diff
@@ -0,0 +1,53 @@
...
+    _dialogService.registerDialogListener(_showDialog);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return widget.child;
+  }
+
+  void _showDialog(DialogRequest request) {
+    var isConfirmationDialog = request.cancelText != null;
```
holtzmak: Would prefer optional `cancelText` in `DialogRequest` than check if null.

app/lib/humane_transport_app.dart
```diff
@@ -1,15 +1,22 @@
...
 import 'core/services/service_locator.dart';
+import 'core/utilities/dialog_manager.dart';
 
 class HumaneTransportApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
       onGenerateRoute: AppRouteGenerator.onGenerateRoute,
       navigatorKey: locator<NavigationService>().navigationKey,
+      builder: (context, child) => Navigator(
```
holtzmak, line 19:
```suggestion
      // Add link to FilledStack page that explains why this needs to be wrapped in a Navigator
      builder: (context, child) => Navigator(
```

app/test/main_test.dart
```diff
@@ -16,6 +17,7 @@ void main() {
     testWidgets('first route is welcome screen', (WidgetTester tester) async {
       addLazySingletonForTest(testLocator, () => WelcomeViewModel());
       addLazySingletonForTest(testLocator, () => NavigationService());
+      addLazySingletonForTest(testLocator, () => DialogService());
```
holtzmak, line 20: Do we need this? I think the current tests do not use `Sign up` code.

holtzmak (approved):
I have some suggestions where we could do null handling better.

I want to see you add a comment in `humane_transport_app.dart` in particular, please do this before merging.

- [x] Please remove `launchAlertDialog`

app/test/main_test.dart
MummenRider: We need this as I used it in signInViewModel. That is just to show how to use the dialog service. I thought it could be helpful to show at least 1 example showing how to use the service

app/lib/core/utilities/dialog_manager.dart
MummenRider: Will do, thanks!

app/lib/core/services/dialog/dialog_service.dart
MummenRider: Changed it to final

app/lib/core/services/dialog/dialog_service.dart
MummenRider: Will do!

PR #159, opened 2021-02-13
Response class

MummenRider:
Will handle success or failure (e.g. Sign in or sign up)

### DIFF
--- a/app/lib/core/utilities/reponse_manager.dart
+++ b/app/lib/core/utilities/reponse_manager.dart
@@ -1,0 +1,17 @@
+import 'package:app/core/utilities/optional.dart';
+import 'package:flutter/material.dart';
+
+enum ResponseType { success, failure, catastrophic_failure }
+
+class Response<T> {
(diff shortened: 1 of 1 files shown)


app/lib/core/utilities/reponse_manager.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:app/core/utilities/optional.dart';
+import 'package:flutter/material.dart';
+
+enum ResponseType { success, failure, catastrophic_failure }
```
holtzmak: - [x] We should still drop the `catastrophic_failure` and just use success and failure.

So, you should also change `Response.failure` to set the type as `failure` by default like what we did in our last discussion.

holtzmak (approved):
Please make the one change lol.

app/lib/core/utilities/reponse_manager.dart
MummenRider: Will do, thanks

PR #160, opened 2021-02-14
Add in progress screen (and big refactoring)

### LINKED ISSUE #133: Create the In-Progress screen
Blocked by #129 and #127 (I think).

_This screen is VERY SIMILAR to #131 , determine if you can reuse any code for this screen (maybe even make the screen have options for IN-PROGRESS or HISTORY filtering) (this doesn't mean that you should have ternary statements everywhere for the two "states" - be mindful of making it scaleable)_

- [x] Create the in-progress screen that will display all the in-progress records in the local database
- [x]  (I think an additional function is needed to translate local database records to ATR objects for display, so as to reduce coupling between database and history) (This work may be done already in #131.)

_This screen should not show all of each record's information, determine from the prototypes what information should be previewed._

This screen WILL HAVE the ability to select a record and open the editing screen for a single record. This work is NOT a part of this task and is the work of #134

holtzmak:
Closes #133 .

Firstly, if you're reviewing, here's how I best suggest to review:
1. Look at the code comparison, and make comments as usual. Fine-detailed review.
> I'm used to managing my own big PRs, so you are welcome to approach as normal.
2. Pull this branch to your machine, and look at the code as-is without regard to specific changes/commits. The big picture. Your best choice.
> This will let you get the idea of what changed, and you can ask questions about particular changes that confused you and ignore changes less important.

***

In this PR , I have done:
1. Renamed/moved files as per our [Hi-Fi design 2](https://xd.adobe.com/view/9a6f1910-d3ac-4fdf-b051-c2c78510ba32-cf3c/). 
> This means that I changed names like "Ongoing" to "Active". 
> This means I renamed `PdfPreview` to `PDFScreen`, and moved it to `ui/.../widgets/utility` because it is a screen, and a utility screen.
> This means that I moved anything making widgets out of `core` and into `ui` unless it was a navigation thing or a view model thing. This is debatable if you like.
2. Cleaned up some uses of `Future`s where our newest tasks sign-up, login, are involved.
> Should be much easier to start theses tasks with the cleanup, and I have added TODOs to guide people. Matches the changes Clark made in the `Firebase-Api` branch.
3. Played around with `Navigator` and `Dialog` services. I found some things that may or may not be concerning:
> We have one service at any time, this means we should only use that service for navigation.
> We are only able to navigate to a screen without the bottom nav bar if we use the root navigator. See `TestScreenThree` or `ATRPreview` for example -> these show how we can navigate to screens not in the route generator and without bottom nav (this is important for screens that are stateful!!!)
> The `FloatingActionButton` has it's own navigator, that's why `PDFScreen` doesn't have the bottom nav bar despite using root navigator
4. I added the `ActiveScreen` and hooked up previews just like `HistoryScreen`
> More interesting things will occur when I add the editing screen
5. Removed most of the useless `print()`s are replaced with `DialogService` where needed
> Let's not add these print statements now. We'll add a logging service later when the code is cleaned up. You're welcome to add these when debugging, but they will not be allowed into the codebase
6. Removed useless TODOs and replaced them with TODOs that mention specific GitHub tasks

***
Questions I have:

@MummenRider , see `TestScreenThree` and how it is used in the `NewScreen`. 

I found that the way I use `Navigator.push` must be the same for every screen that is not a static screen in the navigator (static screen like history screen, non-static widget/screen like ATRPreview). Does this give you any insight on how to add a new function to navigator service so that we can use navigation service for everything?


MummenRider (approved):
LGTM! As I was refactoring our files before, I anticipated nav service problem might occur. I tried to make this work but having really hard time on doing so. I had a discussion with FilledStack members and they suggested:

"If you want the bottomNavigationBar persistant then you need to wrap that in a view and have all other views inside it. Or extract it navigationBar entirely and call it on all the pages you want to use it. Navigation should have nothing to do with it.

The question about using rootNavigator with the navigationService I can't answer. Honestly I've never used it or had any need of it while using autoRoute. I understand your problem but dont think I can help more than that since I don't use nested shizzle like that.

I have a homeView which has my main appBar and bottomNavigationBar and uses a pageView with my main pages. If I want one of the deep pages to have the navigationBar I just put it in there since I already decoupled the complete navigationBar anyway. Sorry man..

If you want true persistance in ALL pages, just create a mainView with the bar and put everything else in there"

Possible Solution I know that I haven't tried yet:
- Call bottom nav bar to page that uses it. For example: Test 3 screen uses it? then call bottomnavbar

holtzmak:
It looks like the work for refactoring `NavService` is available in the `Firebase-API` branch. I'll pick those changes in the next PR.

PR #161, opened 2021-02-16
Add nav service updates

holtzmak:
In this PR I have:
1. Added test(s) for `NavService`
2. Made things that use the Main Navigator use `NavService`
3. Identified that we have to use `Navigator.of(context)` when using inner navigators because they don't have a service
4. Removed the `Drawer` from useless places (like `WelcomeScreen`)
5. Fixed broken tests from changes

There are no GUI changes in this PR (besides the removal of excess drawers in places they shouldn't have been).

@MummenRider I've noticed that the navigation `Drawer` is not persistent in all pages after `HomeScreen`. This confuses me, because the bottom nav bar is persistent and constructed at the same time as the drawer in `HomeScreen`. Any ideas?

Also, @MummenRider what do you think about adding services for the inner navigators? The current service is just the main navigator.


app/lib/core/services/navigation/nav_service.dart
```diff
@@ -9,7 +11,25 @@ class NavigationService {
     return _navigationKey.currentState.pop();
   }
 
-  Future<dynamic> navigateTo(String routeName) {
-    return _navigationKey.currentState.pushNamed(routeName);
+  Future<dynamic> navigateTo(String routeName, {dynamic arguments}) {
+    return Optional(arguments).isPresent()
+        ? _navigationKey.currentState.pushNamed(routeName, arguments: arguments)
+        : _navigationKey.currentState.pushNamed(routeName);
+  }
```
MummenRider, line 18: Wrapping arguments with curly braces means it's optional already, I'm pretty sure if-else isn't necessary here. This applies to all nav services methods you did.

app/lib/ui/views/new/new_screen.dart
```diff
@@ -39,9 +41,8 @@ class NewScreen extends StatelessWidget {
               Padding(padding: EdgeInsets.all(5.0)),
               Text('Go to test page without bottom nav'),
               RaisedButton(
-                onPressed: () => Navigator.of(context, rootNavigator: true)
-                    .push(MaterialPageRoute(
-                        builder: (context) => TestScreenThree())),
+                onPressed: () =>
+                    _navigationService.navigateTo(TestScreenThree.route),
```
MummenRider: You're not supposed to use nav service in UI layer. Here's what you need to do:
```
class NewScreenViewModel extends BaseViewModel{
final NavigationService _navigationService = locator<NavigationService>();
void navigateToTestScreenThree() => _navigationService.navigateTo(TestScreenThree.route);
}
```
In your view, you can simply do: `model.navigateToTestScreenThree();`
Note: This apply to every screen that will navigate

MummenRider (changes requested):
LGTM! Made suggestion when using navigation service . I'm pretty sure you can use NavigatorService in inner loop too. For each inner screen that requires navigation, you simply create a viewmodel for it. I made suggestion that shows how  to implement it

app/lib/core/services/navigation/nav_service.dart
holtzmak: I think I sent you this before:
```
The following NoSuchMethodError was thrown while handling a gesture:
The method 'pushNamed' was called on null.
Receiver: null
Tried calling: pushNamed<Object>("/A", arguments: null)
```
We still need to handle null case, hence why I cast to optional as an easy way to check without forcing people to add `Optional.empty()` when using this function.

MummenRider: oh okay, I didn't know the full context of that error message. If this fixes the problem, then LGTM!

app/lib/ui/views/welcome/welcome_screen.dart
```diff
@@ -1,38 +1,30 @@
...
-                },
+                onPressed: model.navigateToSignInScreen,
                 child: Text('Sign In'),
               ),
               // TODO: Remove eventually, useful for testing now
               RaisedButton(
-                onPressed: () {
-                  Navigator.of(context).pushNamed(HomeScreen.route);
-                },
+                onPressed: model.navigateToHomeScreen,
```
MummenRider, line 27: I think you forgot to add `()` at the end?

MummenRider (approved):
I'll approve this, but please change it before merging. Thanks!

app/lib/ui/views/welcome/welcome_screen.dart
holtzmak: I did not! This is an uncommon way of specifying the function to call.

This is the same as `() => model.navigateToHomeScreen()`
> So, rather than make an anonymous function to call a function, I just say call this function `model.navigateToHomeScreen`. 

Functions can be pointed to like variables, adding `()` calls it, and `() => {}` declares a function. Since `navigateToHomeScreen` is already declared, I just reference it here.

Issue #162, opened 2021-02-17
Consolidate the Navigators into NavigationService


holtzmak:
Having multiple navigators in our app is working against us when we try to navigate between pages across two navigators. Consolidating them will allow us to:
1. Prevent unnecessary widget rebuilds
2. Reduce navigation confusion, we will just use the one `NavService` everywhere
3. Reduce route confusion, all pages will be in the one `onGenerateRoute`
4. Reduce route complexity, only one route is active at a time with no branching routes

***

In this task, you will need to do the following (each removal will force you to remove the next thing, cannot split the work up):
- [x] Remove the `inner navigators` entirely and move their switch cases into `main navigator` (one `onGenerateRoute`)
- [x] Remove `NavigationViewModel` (this will remove the bottom nav bar, this will come back in #163)
- [x] Remove the bottom `BottomNavigationBar` in `HomeScreen`
- [x] Add a button or something to navigate to `NewScreen`, `ActiveScreen`, `HistoryScreen` from `HomeScreen` (HomeScreen should be [this screen from the Hi-Fi ](https://xd.adobe.com/view/9a6f1910-d3ac-4fdf-b051-c2c78510ba32-cf3c/screen/737037be-20e5-4d30-a0c0-66dc2c7a2edf)) (This step is necessary to hook the inner nav screens back into main nav)

Also:
- [x] Go through the screens using `Navigator.of(context)` and make them use a `ViewModel` which uses `NavigationService`
- [x] Make all ViewModels have their necessary `navigateToScreen()` functions as required by their `Screen`(View)

There exists a `NavigationService` test, this should be updated if any changes to `NavigationService`. Also update any screen tests to use a mock ViewModel.

PR #164, opened 2021-02-17
Update ATRDisplayScreen, NavScreenTest, minor other refactorings

holtzmak:
In this PR I have:
1. Updated `ATRDisplayScreen` to use widgets in the expansion tiles
2. Update the `ATR` object classes to have a `toWidget` function for display purposes
3. Refactoring some classes, like renamed `ATRDisplay` to `ATRDisplayScreen`, because it is indeed a screen
4. Obscured the password text in login screen because that bothered me
5. Added another test to `NavService` that shows how we should use just one nav service in #162 . This test was also the reason why we need #162.

***

Until #162 is done, you will see the following in the build output when you click between bottom nav bar items. The reasons for this are explained in #162:
```
======== Exception caught by widgets library =======================================================
The following _Exception was thrown building Builder(dirty):
Exception: Invalid route: /
```
Seeing this in the build output doesn't break the app or anything.

***

Obscured text:

https://user-images.githubusercontent.com/32527219/108150216-efc67500-7099-11eb-8352-fbf80ab68cba.mov

`ATRDisplayScreen`:

https://user-images.githubusercontent.com/32527219/108150328-200e1380-709a-11eb-82b4-76558f166d6f.mov


app/lib/core/view_models/active_screen_view_model.dart
```diff
@@ -44,4 +46,8 @@ class ActiveScreenViewModel extends BaseViewModel {
     _animalTransportPreviews.clear();
     notifyListeners();
   }
+
+  void navigateToActiveScreen() {
+    _navigationService.pop();
```
holtzmak, line 51: This should be `_navigationService.popUntil(ActiveScreen.route)` but we can't do that until #162.

MummenRider (approved):
LGTM

Issue #134, opened 2021-01-26
Create the single ATR Editing screen


holtzmak:
Blocked by #129, #127, and #133.

- [x] Create the editing screen that displays and makes changeable the selected record from the in-progress screen
- [x]  (I think an additional function is needed to replace the existing database record with the updated one, if it was changed. This may already be a part of the database interface in #130 , but you should add a function like `updateRecord()` if it's not)

_This screen should show all of the record's information._

This task could potentially be difficult, recall that there can be many FWR entries per record. How would you add the ability to add more FWR records to an in-progress ATR? The screen(s) will need the dynamic ability to change parts of an ATR.

Some thoughts:
1. Split the editing into multiple screens
2. Block navigation until the user saves (which will update the ATR model in edit before moving to the next screen (which has not replaced the original yet))
3. Have a "main" display screen, maybe reusing the single-ATR history screen here, that the user returns to after editing some portion. The save button here would call on the database interface to replace the original ATR.

PR #167, opened 2021-02-19
Add beginnings of ATR editing screen

holtzmak:
First part of #134 , with the `ShipperInfoFormField` only and some tests. No connection to `DatabaseService` until #130 is done.

***

In this PR I have:
1. Added the ATR editing screen that can take in an existing ATR to edit
2. The ability to take in an existing ATR means we can start a new ATR with cached information and give that to the editing screen to start
3. Each set of fields of the ATR (ShipperInfo, FWR) is to be a separate accordion menu with it's own save button
4. I've added the `ShipperInfoFormField` for the ShipperInfo so far
4. There exists a submit button that calls the save of all the ATR editing fields, will eventually call on database service to save the completed ATR
5. On submit, a popup appears with submission success, failure, and then closes the ATR screen (prevent further editing on submitted ATR)
5. Tests for the form fields as they are so far (no validation yet, obviously we want to block empty fields in some places and prevent code injection yet)

https://user-images.githubusercontent.com/32527219/108447734-0568a580-7226-11eb-8260-57b8a3edec02.mov

***

There will be more PRs to add each of the ATR fields FWR, Loading vehicle info, etc., tests and their interaction with services.


app/lib/core/services/database/database_service.dart
```diff
@@ -1,21 +1,17 @@
 import 'dart:async';
 
 import 'package:app/core/models/animal_transport_record.dart';
-import 'package:app/core/models/test_animal_transport_record.dart';
+import 'package:app/test/test_animal_transport_record.dart';
 
 class DatabaseService {
   // TODO: #130. Replace with actual DB functions
   Stream<AnimalTransportRecord> getUpdatedCompleteATRs() async* {
-    while (true) {
```
holtzmak, line 9: God rid of this because each push of a new record caused excessive rebuilds when playing around on the app.
We really only need the one push. If you sign out and sign back in, the service is remade and will push out a test record after the time completes.

app/test/ui/views/history/atr_display_screen_test.dart
```diff
@@ -1,14 +1,13 @@
...
 import 'package:app/core/services/navigation/nav_service.dart';
 import 'package:app/core/view_models/history_screen_view_model.dart';
+import 'package:app/test/test_animal_transport_record.dart';
 import 'package:app/ui/views/history/atr_display_screen.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:get_it/get_it.dart';
 import 'package:mockito/mockito.dart';
 
-import '../../../test_lib/test_animal_transport_record.dart';
```
holtzmak, line 10: Goodbye relative imports. Hello `lib/test/`!

MummenRider (approved):
LGTM

PR #168, opened 2021-02-19
Firestore query for user account

MummenRider:
This commit covers query for user's account only.

### DIFF
--- a/app/lib/core/services/database/database_interface.dart
+++ b/app/lib/core/services/database/database_interface.dart
@@ -1,26 +1,23 @@
 import 'package:app/core/models/animal_transport_record.dart';
+import 'package:app/core/models/firestore_user.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 
 abstract class DatabaseInterface {
   // TODO: #130. The return types on these functions should not be dynamic
   // Make it explicit what each function returns, if anything
 
-  // For example, newUser should return Future<void>, we only care if it was a success
-  Future<dynamic> newUser(User newUser);
+  Future<void> newUser(FirestoreUser newUser);
 
-  // and getUser should return Future<User>, not dynamic
-  Future<dynamic> getUser(User getUser);
-
-  Future<dynamic> updatePassword(User user);
+  Future<FirestoreUser> getUser(String userId);
 
   Future<dynamic> newRecord(AnimalTransportRecord newRecord);
 
   Future<dynamic> getRecord(User user);
 
   Future<dynamic> updateTranInfo();
-
+  // We might not need this
   Future<dynamic> updateShipperInfo();
-
+  // We might not need this
   Future<dynamic> updateFwr();
 
   Future<dynamic> updateVehicleLoadingInfo();
--- a/app/lib/core/services/database/database_interface.dart
+++ b/app/lib/core/services/database/database_interface.dart
@@ -14,6 +14,8 @@ abstract class DatabaseInterface {
 
   Future<dynamic> getRecord(User user);
 
+  // TODO: #130. Replace these individual methods with single functions like updateIncompleteATR(), addCompleteATR(), removeIncompleteATR(), and so on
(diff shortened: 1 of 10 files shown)


app/lib/core/services/database/database_interface.dart
```diff
@@ -1,26 +1,23 @@
...
-  Future<dynamic> getUser(User getUser);
-
-  Future<dynamic> updatePassword(User user);
+  Future<FirestoreUser> getUser(String userId);
 
   Future<dynamic> newRecord(AnimalTransportRecord newRecord);
 
   Future<dynamic> getRecord(User user);
 
   Future<dynamic> updateTranInfo();
```
holtzmak, line 19:
```suggestion
  // TODO: #130. Replace these individual methods with single functions like updateIncompleteATR(), addCompleteATR(), removeIncompleteATR(), and so on
  // We don't handle ATRs in pieces ATM
  Future<dynamic> updateTranInfo();
```

app/lib/core/services/database/database_service.dart
```diff
@@ -1,10 +1,19 @@
...
+  final DatabaseInterface interface;
+  DatabaseService(this.interface);
+
+  Future<FirestoreUser> getUser(String userId) async =>
+      interface.getUser(userId);
+
+  Future<void> newUser(FirestoreUser newUser) async =>
+      interface.newUser(newUser);
+
   Stream<AnimalTransportRecord> getUpdatedCompleteATRs() async* {
```
holtzmak, line 20:
```suggestion
  // TODO: #130. Replace these functions with the Firestore provider pattern equivalent
  // Meaning, this function may stay but it needs the "real" implementation
  // where we listen to Firestore for changes and get them
  Stream<AnimalTransportRecord> getUpdatedCompleteATRs() async* {
```

app/lib/core/services/database/firebase_database_interface.dart
```diff
@@ -10,10 +11,29 @@ import 'database_interface.dart';
 // TODO: #130. Replace MySQL and SQLFLite with Firebase Database calls.
 // We also probably need a separate service for FireBaseAuthentication
 // Up to you if these are in the same file
-class FirebaseDatabaseInterface extends DatabaseInterface {
+class FirebaseDatabaseInterface implements DatabaseInterface {
+  final FirebaseFirestore _firestore;
```
holtzmak, line 15:
```suggestion
  // We don't use locator for Firestore because mocking requires the instance to be under out control
  // So, we take in the instance in the constructor for testability
  final FirebaseFirestore _firestore;
```

app/lib/core/services/service_locator.dart
```diff
@@ -14,14 +17,21 @@ import 'navigation/nav_service.dart';
 final locator = GetIt.instance;
 
 void setUpLocator() {
+  locator.registerFactory<DatabaseInterface>(
+      () => FirebaseDatabaseInterface(FirebaseFirestore.instance));
+
+  locator.registerLazySingleton(
```
holtzmak: - [ ] Shouldn't this be?

```suggestion
  locator.registerLazySingleton<DatabaseService>(
```

app/pubspec.yaml
```diff
@@ -23,6 +23,8 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  cloud_firestore: ^0.16.0+1
+  connectivity: ^2.0.2
```
holtzmak, line 27: Not that you should remove this, but it should be removed ASAP when we no longer use connectivity (which should be really soon).
```suggestion
  // TODO: Remove this immediately
  connectivity: ^2.0.2
```

holtzmak (approved):
LGTM! Some suggestions for comments that I thought useful. I think the locator registry of DatabaseService needs to have the type specified (at least for consistency).

app/pubspec.yaml
MummenRider: will do so once everything is completed

PR #169, opened 2021-02-20
Add dynamic form field

holtzmak:
Related to #134 .

In this PR I added two form utility classes:
1. `StringFormField`, meant to be used in other `FormFields`
2. `DynamicStringFormField`, meant to be used in other `FormFields`. A dynamic form editor.

Demo:

https://user-images.githubusercontent.com/32527219/108616373-76e55700-73d2-11eb-813b-26c14fea1f3d.mov

Unfortunately, the dynamic form rebuilds a lot. This is because the text editing controller in each field has an initial text, that, if the widget rebuilds causes it to have the initial text again. The widgets rebuild anytime there is a change, so I've elected to accept a lot of rebuilding for a robust form that saves without being told to.


app/lib/ui/views/active/form_field/multi_string_form_field.dart
```diff
@@ -0,0 +1,80 @@
...
+class _MultiStringFormFieldState extends State<MultiStringFormField> {
+  // TODO: #134. This class should offer validation for non-empty lists, if desired
+  // Meaning that the validation function should have optional non-empty check
+  // if at least one field should exist, and the first item should have no delete button
+  final List<StringFormField> _fields = [];
+
+  List<String> getList() => _fields.map((field) => field.getString()).toList();
+
+  void _deleteField(String field) {
+    setState(() {
```
holtzmak: - [x] Oops, this doesn't work the way it should. I should key all the list items.
https://stackoverflow.com/a/59390995

holtzmak: Will add video after I fix this.

app/ios/Podfile.lock
```diff
@@ -1,4 +1,227 @@
 PODS:
+  - abseil/algorithm (0.20200225.0):
```
MummenRider, line 2: Do you what are these pods for?  I've seen this before as part of my experimentation, but no idea how I got it

MummenRider (approved):
LGTM!

app/ios/Podfile.lock
holtzmak: Yeah, `Abseil` is a C++ library (funny name). https://cocoapods.org/pods/abseil 
 It's definitely required by FirebaseFirestore. https://cocoapods.org/pods/FirebaseFirestore (dependencies on the right side of the page)

holtzmak:
@MummenRider ready for re-review (minor review, see video I added in original comment).

PR #172, opened 2021-02-23
Add, update, and remove queries

MummenRider:
- I added a new model which will serve as an initial state of the document. So initially, the document will look like:
```
{
  "userId": "23512",
  "isComplete": false,
}
```
when 'save' button is clicked in shipperInfo, the document will look like:
```
{
    "userId": "23512",
    "isComplete": false,
    "shipInfo": {
      "shipperName": "Clark",
      "shipperIsAnimalOwner": false,
      "departureLocationId": "123",
      "departureLocationName": "SomePlace",
      "shipperContactInfo": "213",
      "departureAddress": {
        "streetAddress": "Bruce",
        "city": "Regina",
        "provinceOrState": "some state",
        "postalCode": "dfsdf",
        "country": "Canada"
      }
    }
}
```
Once user completed everything, we will update the initalAtr object and set the `isComplete: true`

The behaviour of firestoreDb will look like the video below.

https://user-images.githubusercontent.com/43891568/108789914-0577e680-7541-11eb-8e9e-d55f0ff801d2.mov

- If we follow this, then we might not need the existing AnimalTransportRecord model as we are not passing infos as whole

- Notice that there's no update to queries that uses `set()`, this is because it will add document/ field if it does not exist. Otherwise, it'll just update.

- As for adding newAtr, I used `add()` so that firestore generates unique Id for us. To update this, I create a method that will specifically update the atr. The update function can be used when changing the isComplete field to true. 
- The atrId will be used to determine which atr will be updated.  This idea is based on [FilledStack](https://www.youtube.com/watch?v=1PhAPWzGaM4&t=1257s&ab_channel=FilledStacks) tutorial. Specifically, watch the video starting from 18:40


holtzmak:
@MummenRider 
> If we follow this, then we might not need the existing AnimalTransportRecord model as we are not passing infos as whole

We still bundle the model for display. Easier to pass 1 object around than 9.

app/lib/core/services/database/firebase_database_interface.dart
```diff
@@ -1,124 +1,105 @@
...
+  @override
+  Future<void> setLoadingVehicleInfo(
+          String atrId, LoadingVehicleInfo vehicleInfo) async =>
+      _firestore
+          .collection('atr')
+          .doc(atrId)
+          .set(vehicleInfo.toJSON(), SetOptions(merge: true));
+
+  @override
+  Future<void> updateAtr(InitialAtr atr) async =>
```
holtzmak: - [x] Should this take the full ATR model object eventually?

app/lib/core/models/initial_atr.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter/material.dart';
+
+@immutable
+class InitialAtr {
```
holtzmak: - [x] I think a better way to use this class would be to name it `AtrIdentifier`. Because it identifies the ATR, so then the `AnimalTransportRecord` model object will have this object in it instread of separate fields for `author` and `atrID`

MummenRider:
> @MummenRider
> 
> > If we follow this, then we might not need the existing AnimalTransportRecord model as we are not passing infos as whole
> 
> We still bundle the model for display. Easier to pass 1 object around than 9.

Correct me if I'm wrong, but I'm pretty sure you can grab everything and display it  just by getting the atrId.
We can keep that model for sure.

app/lib/core/services/database/database_interface.dart
```diff
@@ -10,23 +17,29 @@ abstract class DatabaseInterface {
 
   Future<FirestoreUser> getUser(String userId);
 
-  Future<void> newRecord(AnimalTransportRecord newRecord);
+  Future<void> saveNewAtr(InitialAtr atr);
 
-  Future<dynamic> getRecord(User user);
+  Future<void> updateAtr(InitialAtr atr);
```
holtzmak: What confuses me about this is that you are indeed giving the identifier to update, but the `InitialAtr` has no information that updates besides` user ID, atrID, is complete`. 

We could put `AnimalTransportRecord` here, but I think what you're getting at in the PR is that we don't need to hold the temporary ATR and instead just it's parts.

- [x] I think this should not exist then, if we use the other functions `setField` piecemeal.

But, I think we still need to use `AnimalTransportRecord` for active (in-progress) ATRs. In editing the sections of the ATR we can just hook up its specific `setField` function for updating (your work is very good for that!). 

However, we still allow for editing the whole ATR at once (even if we change the design to split sections into pages). Hence, the collective ATR object still has use. Consider also when the app is brand new, if the shipper has an ATR in progress we still pull the whole record from the database.

holtzmak: This is very open to discussion though, talk to me on group chat.

app/lib/core/models/initial_atr.dart
holtzmak:
```
AnimalTransportRecord {
   final AtrIdentifier identifier;
   // ...
}
```

app/lib/core/services/database/firebase_database_interface.dart
```diff
@@ -1,124 +1,105 @@
...
 import 'database_interface.dart';
 
 // TODO: #130. Replace MySQL and SQLFLite with Firebase Database calls.
 // We also probably need a separate service for FireBaseAuthentication
 // Up to you if these are in the same file
 class FirebaseDatabaseInterface implements DatabaseInterface {
   final FirebaseFirestore _firestore;
-
-  var networkStatus;
-  MySqlConnection dbConnect;
```
holtzmak, line 18: - [x] Can we please remove connectivity plugin from `pubspec.yaml`?

app/lib/core/services/database/firebase_database_interface.dart
```diff
@@ -1,124 +1,105 @@
...
+  Future<void> newUser(FirestoreUser newUser) async =>
+      _firestore.collection('users').doc(newUser.userId).set(newUser.toJSON());
 
-  Future<dynamic> updateAckInfo() async {
-    dbConnect = await connectRemote();
-    var updateAck = await dbConnect.query('''DATABASE QUERY''');
-    dbConnect.close();
-    return updateAck;
+  @override
+  Future getRecord(User user) {
```
holtzmak, line 33: I imagine @SanaKhan1997 is working on the group of functions for fetching in-progress ATRs, ATRs that match search critera, ATRs from the last month, etc.?

app/test/core/services/database/db_service_test.dart
```diff
@@ -59,15 +58,37 @@ void main() {
...
           .thenReturn(mockCollectionReference);
-      await dbService.newRecord(userModel);
-      verify(mockCollectionReference.add(userModel.toJSON()));
+      await dbService.saveNewAtr(userModel);
+      verify(mockCollectionReference.add(userModel.toJSON())).called(1);
+    });
+
+    test('should delete atr to firestore', () async {
+      final userModel = InitialAtr.fromJSON(_fakeInitialAtr, '123');
+      when(mockFirestoreInstance.collection(any))
```
holtzmak, line 74: Just FYI, you can extract duplicate `when()` statements you need to a helper function. I think it should work the same way (at least it does in Java, Kotlin)

app/test/core/services/database/test_data.dart
```diff
@@ -10,6 +10,14 @@ Map<String, dynamic> fakeData() {
   return _fakeResponse;
 }
 
+Map<String, dynamic> fakeInitialAtr() {
```
holtzmak, line 13: - [x] Please move `test_data`, `test_helper` with the mocks into the `lib/test/` folder.

I was helping Mansi with `AuthenticationService` tests, which uses these files. Without being under `lib` the files have relative imports which are annoying.

holtzmak (changes requested):
Mostly questions I'd like answered about where I think you're going with this.

I requested some minor changes and will review again after the discussion I expect from my questions.

EDIT: Are you sure this closes #130 ? I'm pretty sure more record fetching functions need to be implemented

app/lib/core/services/database/firebase_database_interface.dart
MummenRider: This updates only the InitialAtr. So, userId and isComplete.  The code with SetOptions will update the objects. You can see this on the video, I used the SetOptions to update.

app/lib/core/models/initial_atr.dart
MummenRider: What is your plan for the rest of the fields on your example? We need a class that will push specifically userId and isComplete fields initially. If you plan to add more here, I don't think it'll work.

app/lib/core/services/database/firebase_database_interface.dart
holtzmak: I see, the name `updateATR` was misleading to me because I thought it updated the whole ATR or would eventually. I see now this is your intended way to update `isComplete`.

app/lib/core/models/initial_atr.dart
holtzmak: We would first use `InitialAtr`, then when we make the rest of the `ATR` we give it the `InitialAtr` and start adding the other fields as we go. That would mean that the ATR should have `Optional` ShipperInfo, etc. until the next section is added.

app/lib/core/services/database/database_interface.dart
holtzmak: Resolved by [other comment](https://github.com/holtzmak/Humane-Transport/pull/172#discussion_r581157964).

app/lib/core/models/initial_atr.dart
holtzmak: We'll discuss the full workflow of creating a new ATR, editing it, saving it, repeating those two previous steps, and completing it, either in the next scrum or backlog refinement. 

This discussion goes beyond GitHub comments and is the result of confusion as to how code communicates between modules. Same with [this comment](https://github.com/holtzmak/Humane-Transport/pull/172#issuecomment-783908390).

app/lib/core/models/atr_identifier.dart
```diff
@@ -0,0 +1,38 @@
...
+    this.atrDocumentId,
+    this.isComplete = false,
+  });
+
+  factory AtrIdentifier.fromJSON(
+      Map<String, dynamic> json, String atrDocumentId) {
+    return AtrIdentifier(
+      userId: json['userId'],
+      isComplete: json['isComplete'],
+      atrDocumentId: atrDocumentId,
```
holtzmak, line 20: Isn't the document ID in the JSON?

holtzmak (approved):
LGTM! But, are we not using ATR ID in the database? Can you tell me again why we only use the ID in the code?

app/lib/core/models/atr_identifier.dart
MummenRider: We don't need to pass atrId because it's sole purpose is to identify which atr needs to be updated. We use this behind the scene. This idea is based on [FilledStack](https://www.youtube.com/watch?v=1PhAPWzGaM4&t=1257s&ab_channel=FilledStacks) tutorial. If you'd like to see it, start on 18:40.
In filledstack's tutorial he named this as `documentId`. In our case I named it as `atrDocumentId`

PR #174, opened 2021-02-26
Add transport info to editing screen

holtzmak:
Related to #134 .
In this PR I have:
1. Made all our form fields conform to the same `Form` ft. callback configuration
> We need this configuration with lots of save callbacks to underlying ATREditingScreen so widget rebuilds don't interrupt editing. The callbacks are simple so they don't consume too much rebuilds. Unfortunately, I have to shrink wrap our ListViewBuilder in the Column so the most expensive rebuild is the dynamic string list. More dynamic lists incoming.
2. Added Transporter Info editing to the ATREditing screen
3. Squashed some bugs like the `DateTime.now()`s which can cause some bad comparison behaviour in tests

***

What the changes look like:

https://user-images.githubusercontent.com/32527219/109249858-22652180-77ae-11eb-89d4-606e901f6a27.mov


app/lib/ui/views/active/form_field/address_form_field.dart
```diff
@@ -1,88 +1,99 @@
...
-    ]);
+    return Form(
+      key: widget.formKey,
+      child: Column(children: [
+        StringFormField(
+            initial: _street,
+            title: "Street",
+            onSaved: (String changed) {
+              _street = changed;
+              _saveAll();
```
holtzmak, line 61: You may be thinking "Woah, that's a lotta saves!" but these saves go to temporary objects that, only once the person clicks submit or exits the editing screen do the changes get communicated to the database.

app/lib/ui/views/active/form_field/shipper_info_form_field.dart
```diff
@@ -1,112 +1,123 @@
...
+            onSaved: (String changed) {
+              _name = changed;
+              _saveAll();
+            },
+            onDelete: Optional.empty()),
+        ListTile(
+            visualDensity: VisualDensity(horizontal: 0, vertical: -2),
+            title: Text(
+                "The shipper is the owner of the animals loaded in the vehicle?"),
+            subtitle: DropdownButtonFormField(
```
holtzmak, line 79: I can't seem to extract this out as the `onChanged` complains the `(String changed) ` isn't a `(dynamic changed)` 🤷 .

app/lib/ui/views/active/form_field/transporter_info_form_field.dart
```diff
@@ -0,0 +1,216 @@
...
+      {Key key, @required this.initialInfo, @required this.onSaved})
+      : super(key: key);
+
+  @override
+  _TransporterInfoFormFieldState createState() =>
+      _TransporterInfoFormFieldState();
+}
+
+class _TransporterInfoFormFieldState extends State<TransporterInfoFormField> {
+  String _companyName;
```
holtzmak, line 29: Ugly as this is, this is the same as the ViewModel's I'm familiar with in `TornadoFx`. Gotta have that temporary object bits for caching.

app/test/ui/views/active/atr_editing_screen_test.dart
```diff
@@ -30,8 +30,9 @@ void main() {
    */
   void verifyShipperInfoIsShown(ShipperInfo infoExpected) {
     expect(find.text(infoExpected.shipperName), findsOneWidget);
-    expect(find.text(infoExpected.shipperIsAnimalOwner ? "Yes" : "No"),
-        findsOneWidget);
+    // TODO: will have to verify this another way as there is a number of widgets with Yes/N0
```
holtzmak, line 33: Likely involves keying the objects.

MummenRider (approved):
LGTM!

PR #179, opened 2021-02-27
Add equality operator overloads and bring back toString() for tests

holtzmak:
In this PR I have done:
1. Added back the `toString()` methods for tests. Can also be used when generating the PDF until we format it.
2. Added equality and hashcode overloads for tests. These might be a little off, but they will definitely be caught in tests because of `toString()`.
3. Made a bunch of functions into dart syntactic sugar getters


SanaKhan1997 (approved):
Look good! thanks

PR #182, opened 2021-02-28
Add ack and contingency info editing screen

### LINKED ISSUE #134: Create the single ATR Editing screen

holtzmak:
Closes #134 !!!!!

In this PR I have added:
1. The placeholder for ack info in the editing screen. That's probably where we'll launch image picker overlay
2. `ContingencyPlanInfoFormField` and it's sub forms

***

Looks like:

https://user-images.githubusercontent.com/32527219/109409783-15396580-795b-11eb-86d2-6c554f25aaf8.mp4

### DIFF
--- a/app/lib/core/models/contingency_plan_info.dart
+++ b/app/lib/core/models/contingency_plan_info.dart
@@ -94,7 +94,7 @@ class ContingencyPlanInfo {
   String toString() =>
       '''Goal Statement (company’s goal and purpose of the plan i.e avoid animal suffering): $goalStatement
       Communication Plan (who should be contacted and who will initiate or permit the process?): $communicationPlan
-      Crisis contacts and links( general helpline, industry related links and websites): ${_crisisContacts.join(',')}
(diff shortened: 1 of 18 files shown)


SanaKhan1997 (approved):
LGTM

PR #183, opened 2021-02-28
Stream and fetch methods

### LINKED ISSUE #130: Create the Firebase Database interface implementation

SanaKhan1997:


app/lib/core/models/acknowledgement_info.dart
```diff
@@ -28,9 +28,12 @@ class AcknowledgementInfo {
   }
 
   AcknowledgementInfo.fromJSON(Map<String, dynamic> json)
-      : shipperAck = File(json['shipperAck']),
-        transporterAck = File(json['transporterAck']),
-        receiverAck = File(json['receiverAck']);
+      // : shipperAck = File(json['shipperAck']),
```
holtzmak, line 31: We'll go back and fix this come Sprint 9 when we bring in the image picker.

app/lib/core/models/contingency_plan_info.dart
```diff
@@ -194,7 +194,7 @@ class ContingencyPlanEvent {
   }
 
   ContingencyPlanEvent.fromJSON(Map<String, dynamic> json)
-      : eventDateAndTime = DateTime.parse(json['eventDateAndTime']),
+      : eventDateAndTime = json['eventDateAndTime'].toDate(),
```
holtzmak, line 197: @holtzmak to test this on iPhone. We'll probably need a switch case as per https://stackoverflow.com/a/55398555

holtzmak (approved):
LGTM. I was with you while you were making this :)

PR #185, opened 2021-03-02
Sign in page

### LINKED ISSUE #153: Create the Login Screen
Blocked by #151.

- [x] Create the Login screen as per our hi-fi design

A few things to consider:
1. You will be making the GUI screen and the controller (the `ViewModel` which talks to the service below)
2. You may need to make an `AuthenticationService` that talks to the database for logging in (this could already have been done in #152 , go check)
3. You need to handle errors that could occur if talking to the database failed

Fortunately, @MummenRider has already investigated the work needed to make this screen. Talk to him and see what code you can salvage from his work on #95.

_You may also consider additional things like: two factor authentication offered by firebase, sending a confirmation email or something. Do these now if easy, else leave for a later Sprint._

mansiPaTeLL:
@holtzmak @MummenRider @SanaKhan1997 Sign-in page PR is ready to review. Thanks :)

### DIFF
--- a/app/lib/ui/views/signin/sign_in_screen.dart
+++ b/app/lib/ui/views/signin/sign_in_screen.dart
@@ -25,7 +25,6 @@ class _SignInScreenState extends State<SignInScreen> {
   final TextEditingController emailController = TextEditingController();
   final TextEditingController passwordController = TextEditingController();
 
-
   void dispose() {
     emailController.dispose();
     passwordController.dispose();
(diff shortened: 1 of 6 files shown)


app/lib/ui/views/signin/sign_in_screen.dart
```diff
@@ -42,6 +68,8 @@ class SignInScreen extends StatelessWidget {
                     ),
                   ),
                   TextFormField(
+                    key: ObjectKey("Email"),
```
holtzmak: When we do system tests, we'll need to pay attention to whether sign in screen and sign up screen exist in the widget tree at the same time. Since they have the same object keys, grabbing an element by the object key `Email` would grab both or we'll have an error at runtime that two widgets exist with the same key. I think keys are global.

app/lib/ui/views/signin/sign_in_screen.dart
```diff
@@ -88,17 +120,26 @@ class SignInScreen extends StatelessWidget {
                   SizedBox(
                     height: 100.0,
                   ),
+                  TextButton(
+                    child: Text('Do not have an account? Sign Up Here'),
+                    onPressed: () => widget._navigationService.navigateTo(SignUpScreen.route),
```
holtzmak: I think to solve my problem in a [previous comment](https://github.com/holtzmak/Humane-Transport/pull/185#discussion_r585763870), we should do `navigationService.navigateBackTo(homescreen)` then `navServ.push(login)`. Cleaner navigation than causing a loop between welcome, login, and sign up.

- [x] Do pop until, then push

I can also do this change in #162 .

app/test/ui/views/signin/sign_in_screen_test.dart
```diff
@@ -0,0 +1,116 @@
...
+          // Non-standard animation time due to decoration
+          await tester.pump(Duration(milliseconds: 1));
+
+          verify(mockSignInViewModel.signIn(
+              email: "email@email.com",
+              password: "Test pas",
+              )).called(1);
+        });
+
+    testWidgets('shows register button', (WidgetTester tester) async {
```
holtzmak, line 60:
```suggestion
    testWidgets('shows sign in button', (WidgetTester tester) async {
```

app/test/ui/views/signin/sign_in_screen_test.dart
```diff
@@ -0,0 +1,116 @@
...
+          find.widgetWithText(TextButton, 'Do not have an account? Sign Up Here');
+          await tester.ensureVisible(dontHaveAnAccountButtonFinder);
+          expect(dontHaveAnAccountButtonFinder, findsOneWidget);
+          await tester.tap(dontHaveAnAccountButtonFinder);
+          // Non-standard animation time due to decoration
+          await tester.pump(Duration(milliseconds: 1));
+          verify(mockNavigationService.navigateTo(SignUpScreen.route)).called(1);
+        });
+
+    testWidgets('empty field validators appears on empty fields',
```
holtzmak, line 93: We'll make a task for a `ValidationService` soon. You won't need to write tests for validators in widget tests once that is done, because `ValidationService` will use simple unit tests.

holtzmak (changes requested):
Do you need help rebasing this branch? Let me know, it looks like it's not passing checks.

app/lib/ui/views/signin/sign_in_screen.dart
holtzmak: This is part of #162, don't do this.

holtzmak (approved):
LGTM

PR #193, opened 2021-03-07
Account screen

### LINKED ISSUE #155: Create the Account screen
Blocked by #151 , #152 , #153 .

- [x] Create the account information screen (the one that shows your Transporter name, email, contact phone #, when logged in) as per our Hi-Fi design

This task includes the GUI screen and the controller (the `ViewModel`).

_This task does not involve adding the feature for editing the information, that is another task_

### LINKED ISSUE #156: Create the Account editing screen
Blocked by #155 .

- [x] Add the screen for editing account information

You may refactor the existing account screen with a simple editing button and make the fields changeable and saveable; or you may make a whole new screen. If a new screen, you will need a new controller (the `ViewModel`).

MummenRider:
Note: Testing for this task will be done on separated PR.

### DIFF
--- a/app/lib/core/services/database/database_interface.dart
+++ b/app/lib/core/services/database/database_interface.dart
@@ -12,6 +12,8 @@ abstract class DatabaseInterface {
 
   Future<void> updateAtr(AnimalTransportRecord atr);
 
+  Future<void> updateTransporter(Transporter transporter);
+
   Future<void> removeAtr(String atrId);
 
   Future<List<AnimalTransportRecord>> getCompleteRecords(String userId);
@@ -22,5 +24,5 @@ abstract class DatabaseInterface {
 
   Stream<List<AnimalTransportRecord>> getUpdatedActiveATRs(String userId);
 
-  Stream<Transporter> getUserInRealTime(String userId);
+  Stream<Transporter> getUpdatedTransporter(String userId);
 }
--- a/app/lib/core/services/database/database_interface.dart
+++ b/app/lib/core/services/database/database_interface.dart
@@ -6,13 +6,15 @@ abstract class DatabaseInterface {
 
   Future<Transporter> getTransporter(String userId);
 
+  Future<void> updateTransporter(Transporter transporter);
+
   Future<void> removeTransporter(String userId);
+
+  Stream<Transporter> getUpdatedTransporter(String userId);
 
   Future<AnimalTransportRecord> setNewAtr(String userId);
 
   Future<void> updateAtr(AnimalTransportRecord atr);
-
-  Future<void> updateTransporter(Transporter transporter);
 
   Future<void> removeAtr(String atrId);
 
@@ -23,6 +25,4 @@ abstract class DatabaseInterface {
   Stream<List<AnimalTransportRecord>> getUpdatedCompleteATRs(String userId);
 
   Stream<List<AnimalTransportRecord>> getUpdatedActiveATRs(String userId);
-
-  Stream<Transporter> getUpdatedTransporter(String userId);
 }
--- a/app/lib/core/services/database/database_interface.dart
+++ b/app/lib/core/services/database/database_interface.dart
@@ -21,4 +21,6 @@ abstract class DatabaseInterface {
   Stream<List<AnimalTransportRecord>> getUpdatedCompleteATRs(String userId);
 
   Stream<List<AnimalTransportRecord>> getUpdatedActiveATRs(String userId);
+
+  Stream<Transporter> getUserInRealTime(String userId);
 }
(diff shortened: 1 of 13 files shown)


app/lib/core/services/database/database_interface.dart
```diff
@@ -21,4 +21,6 @@ abstract class DatabaseInterface {
   Stream<List<AnimalTransportRecord>> getUpdatedCompleteATRs(String userId);
 
   Stream<List<AnimalTransportRecord>> getUpdatedActiveATRs(String userId);
+
+  Stream<Transporter> getUserInRealTime(String userId);
```
holtzmak: Since the User and Transporter are separate entities, this would be better as:
```suggestion
  Stream<Transporter> getUpdatedTransporter(String userId);
```
The database interface is separate from the logic of the app.

app/lib/core/view_models/account_edit_view_model.dart
```diff
@@ -0,0 +1,54 @@
...
+    setState(ViewState.Busy);
+    var updateTransporterAccount = Transporter(
+      firstName: firstName,
+      isAdmin: _account.isAdmin,
+      lastName: lastName,
+      userEmailAddress: userEmailAddress,
+      userId: _account.userId,
+      userPhoneNumber: userPhoneNumber,
+    );
+    _databaseService.addTransporter(updateTransporterAccount).then((_) =>
```
holtzmak: The `addTransporter` function is not an update. This will create duplicates of transporters in the database.

- [x] Fix by adding `updateTransporter` like the `updateAtr` function in DBS

app/lib/ui/views/account/account_editing_screen.dart
```diff
@@ -0,0 +1,187 @@
...
+import 'package:app/ui/widgets/utility/template_base_view_model.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_multi_formatter/flutter_multi_formatter.dart';
+
+class AccountEditingScreen extends StatefulWidget {
+  static const route = '/accountEditingScreen';
+  final formKey = GlobalKey<FormState>();
+  final Transporter account;
+
+  AccountEditingScreen({Key key, this.account}) : super(key: key);
```
holtzmak: - [x] Account should be required. No null accounts! This will change your null checks in the state below

_Your null checks below are also inconsistent. You check if the fields are null but then pass a possibly null account to the model. This is resolved by doing the above._

app/lib/core/view_models/account_screen_view_model.dart
```diff
@@ -0,0 +1,31 @@
...
+class AccountScreenViewModel extends BaseViewModel {
+  final NavigationService _navigationService = locator<NavigationService>();
+  final DatabaseService _databaseService = locator<DatabaseService>();
+  final AuthenticationService _authenticationService =
+      locator<AuthenticationService>();
+
+  Transporter _transporter;
+  Transporter get transporter => _transporter;
+
+  void loadTransporterInfo() {
```
holtzmak: - [x] This stream subscription, `getUserInRealTime` needs to be disposed of like `TextEditingControllers`

See the HistoryScreenViewModel for example.

app/lib/ui/views/account/account_editing_screen.dart
```diff
@@ -0,0 +1,187 @@
+import 'package:app/core/models/transporter.dart';
+import 'package:app/core/view_models/account_edit_view_model.dart';
+import 'package:app/ui/common/view_state.dart';
+import 'package:app/ui/widgets/utility/busy_overlay_screen.dart';
+import 'package:app/ui/widgets/utility/template_base_view_model.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_multi_formatter/flutter_multi_formatter.dart';
```
holtzmak, line 7: I don't see this being used anywhere?

holtzmak (changes requested):
Some quick changes I'd like to see: otherwise you could have unsafe code (null-wise or memory-leaks).

I have further questions I'll discuss with you by DM.

app/lib/ui/views/account/account_editing_screen.dart
MummenRider: multi fromatter is used to format the phone number. For instance, '+1 (306) 123 456'. This format is based on our prototype design

holtzmak: Oh, my bad. I see it now. We should use this in the sign up screen as well

MummenRider:
I don't know what happened to '23 changed files' but I am pretty sure I did not change 23 files in this commit. The 23 file changes happened when I tried to resolve merge conflicts. It pass all the checks, so I assume nothing is wrong @holtzmak ?

holtzmak (approved):
@MummenRider I rebased your branch to fix whatever happened.

I reviewed as I went through it, and added the listening for user auth changes when launching the account screen (should prevent people from changing records when not authenticated. The database rules catch this, too).

LGTM! Especially if you would add some unit tests later.

PR #194, opened 2021-03-07
Update AuthService to provide current user changes. Check if user is logged in on some screens.

### LINKED ISSUE #186: Revise login and Authentication service so the current logged in user can be used
Right now, when the user logs in, we don't actually watch for changes in credential status: we aren't providing a way to get changes to the current user in AuthService. We need to be listening for changes so we can prevent unauthorized users from receiving information not meant for them, prevent database interface calls when not logged in, etc.

- [x] Update AuthenticationService to provide the current auth user

This would require using something alike the following, but more complex for our application.
```
FirebaseAuth.instance
  .authStateChanges()
  .listen((User user) {
    if (user == null) {
      print('User is currently signed out!');
    } else {
      print('User is signed in!');
    }
  });
  ```

holtzmak:
Closes #186 .

This PR:
1. Addresses a few bugs with the last PR I made
> On signout, the living viewmodels scrap listening to atr updates and user changes before destruction now.
2. Made `AuthService` provide just `FirebaseAuth` changes.
> @MummenRider 's open PR #193 has Transporter changes through the database, which is correct and the way I should have done it.
3. Made `ViewModel`s listen to `AuthService` for active user changes where needed
> Changed any buttons with the one-time get of logged in user, and made any dynamic listening screens stop when user logs out (this prevents unwanted changes due to some bug that keeps the user on the screen when logged out)

There are still unfortunately some bugs due to navigation. Particularly, when the user closes the app and opens it again, if they are logged in we do indeed skip to the home page, but if they sign out after, the app gets mad because the screen to pop to lost it's state (`SignInScreen`).

### DIFF
--- a/app/lib/core/services/authentication/auth_service.dart
+++ b/app/lib/core/services/authentication/auth_service.dart
@@ -13,21 +13,34 @@ class AuthenticationService {
   final databaseService = locator<DatabaseService>();
   final FirebaseAuth firebaseAuth;
 
+  // Not disposed of in destructor because AuthService goes out of scope when
+  // app is destroyed. The disposal is handled by garbage cleanup. Also, Dart
+  // does not define destructors https://github.com/dart-lang/sdk/issues/3691
+  StreamSubscription<User> _authStateChanges;
+
   // The logged in user information is the same as the Transporter information
   // but separate for Firebase Authentication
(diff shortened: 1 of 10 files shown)


SanaKhan1997 (approved):
LGTM

PR #198, opened 2021-03-08
Consolidate navigation

### LINKED ISSUE #162: Consolidate the Navigators into NavigationService

holtzmak:
Closes #162 .

This PR addresses the work outlined in #162. We no longer see exceptions when navigating the app now.

Also fixes the bug with login from #194 .

> You may ask yourself, "don't the active and history screens rebuild a bunch? Won't this cause more read/writes in the database"? I'm pretty sure they do not! The database service is static, while the user is logged in the streams pull in updates to the cache as necessary, and when we rebuild the view models, we simply reattach to the cache! The only time we pull on demand is the first time the widgets are built after logging in.

***

Changes look like so (got this capture before rebasing with Mansi's changes, but they are the same)

https://user-images.githubusercontent.com/32527219/110394100-72a37580-8031-11eb-8727-8fcdb8bc9abe.mp4

https://user-images.githubusercontent.com/32527219/110394188-9e266000-8031-11eb-91ef-9f03e1faadc5.mp4

https://user-images.githubusercontent.com/32527219/110394207-a8485e80-8031-11eb-9a31-78df9fb2f764.mp4


MummenRider (approved):
LGTM

Issue #214, opened 2021-03-10
Update application GUI with a proper color theme


holtzmak:
- [x] Update the GUI and it's components to look better

mansiPaTeLL:
@holtzmak @MummenRider Sana and I worked on GUI colour theme. Please review the designs below - 

**Color Scheme#1**
![Screen Shot 2021-03-11 at 5 01 30 PM](https://user-images.githubusercontent.com/44125953/110867299-1e480200-828c-11eb-9330-c73d324a2238.png)

**Color Scheme#2**
![Screen Shot 2021-03-11 at 5 02 25 PM](https://user-images.githubusercontent.com/44125953/110867332-21db8900-828c-11eb-812d-f190303d477a.png)

holtzmak:
Gotta say no to design 2 because one of the other capstone projects has the exact same color scheme. However, I like that it's more white with color accents.

I think we should combine Design 1 and 2: I like the scheme of 1, but we should swap the brown and white (so, for example the Home page buttons are white on a brown background).

mansiPaTeLL:
@holtzmak @MummenRider We have also selected some other schemes as well. We have already used scheme#1, and scheme#2. 
![unknown](https://user-images.githubusercontent.com/44125953/110867910-310f0680-828d-11eb-847f-6dfe823e2c35.png)

mansiPaTeLL:
@SanaKhan1997 I am going to add color schemes into sign up, login and password reset.

SanaKhan1997:
> @SanaKhan1997 I am going to add color schemes into sign up, login and password reset.

I've already applied color scheme to whole app. Will create PR soon

Issue #205, opened 2021-03-10
Add validation to the ATREditingScreen and forms


holtzmak:
This could be a big task.

- [x] Extract existing validators into a sort of `ValidationService`
- [x] Use the validators in the form fields and dynamic form fields
- [x] Since the dynamic form fields are weird, the current method of `formKey.currentState.save()` doesn't work on them. Refactor the dynamic form fields so we can do this.

Refactoring likely involves doing something like this [example](https://stackoverflow.com/questions/54848639/flutter-custom-formfield-validate-and-save-methods-are-not-called). Particularly, the form fields in the dynamic form (`StringFormField`, `IntFormField`) should extend `FormField` and not `StatefulWidget`

PR #220, opened 2021-03-14
Update form fields for validation

holtzmak:
The big part of #205. Some TODOs are mistakenly 136 or 134. I'll being going through app TODOs and assigning them next PR.

In this PR I have done:
1. Update all form fields to have some form of validation
2. Update the form fields to have the outlined box decoration. Maybe this is possible using app theme.
3. Update the editing screen so saving the ATR when backing out (not hitting save or submit buttons) saves as-is
4. Update the save and submit buttons to validate their respective forms
> Validation, when called outside the state you're validating, will not cause the red text to appear. The save button in that widget will cause the red text to appear. 
>
> So, when hitting submit, we can't make the red text appear, but we can tell the transporter which section needs to be looked at again. I'll be doing this in the next PR.
5. Update `DynamicFormField` to be less generically confusing, also dropped the expensive `ListView.builder` shrink wrap
6. Updated tests accordingly
7. Fixed a hidden bug with the `TimeOfDay` type and the database. It should be `DateTime` where we ignore the date. Changed accordingly

I'll be leaving further GUI changes to the form fields to #214 (like making the menus larger, adding white text to the save and submit buttons, because they're black now as seen in the video (hard to see))

***

Changes look like:

https://user-images.githubusercontent.com/32527219/111082800-c4d01500-84cf-11eb-97b5-0978c72dff2b.mp4

https://user-images.githubusercontent.com/32527219/111082801-c7326f00-84cf-11eb-8c34-9d677277ff15.mp4

https://user-images.githubusercontent.com/32527219/111082807-cbf72300-84cf-11eb-9191-4033c1bd2d1e.mp4

### DIFF
--- a/app/lib/core/models/contingency_plan_info.dart
+++ b/app/lib/core/models/contingency_plan_info.dart
@@ -278,7 +278,7 @@ class ContingencyPlanEvent {
 
 @immutable
 class ContingencyActivity {
-  final TimeOfDay time;
+  final DateTime time;
   final String personContacted;
   final String methodOfContact;
   final String instructionsGiven;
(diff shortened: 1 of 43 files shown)


MummenRider (approved):
All changes here LGTM

mansiPaTeLL:
@holtzmak  LGTM.

mansiPaTeLL (approved):
@holtzmak LGTM

PR #222, opened 2021-03-15
Added query to upload images to firebase_storage

### LINKED ISSUE #215: Investigate and create the ability to save images in Cloud Storage for Firebase
- [x] Set up Cloud Storage and our database so we can save transporter acknowledgements with the ATR
- [x] Update DatabaseService accordingly

[From preliminary investigation](https://firebase.google.com/docs/storage) it looks like we upload the image to cloud storage and then use it's link in our cloud firestore.

This means we fetch and save records with the image URL/link, then read that link and contact cloud storage to fetch it to show in app. Similarly, we contact cloud storage first to save the image, then use the given link in cloud firestore.

MummenRider:
Added a query that will create the ability to upload images to firebase storage. Possible tasks that could be open is the ViewModel that will use the service itself. I also added a really useful plugin called permission_handler where it can be use in the viewmodel by checking the permission.

[Link](https://www.youtube.com/watch?v=KmyhznxYQ70&t=1096s&ab_channel=AndyJulow) to the tutorial I followed for this task.

### DIFF
--- a/app/lib/core/services/database/database_interface.dart
+++ b/app/lib/core/services/database/database_interface.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:app/core/models/animal_transport_record.dart';
 import 'package:app/core/models/transporter.dart';
 
@@ -17,6 +19,8 @@ abstract class DatabaseInterface {
   Future<void> updateAtr(AnimalTransportRecord atr);
 
   Future<void> removeAtr(String atrId);
+
+  Future<String> uploadAtrImage(File file, String fileName);
 
   Future<List<AnimalTransportRecord>> getCompleteRecords(String userId);
 
--- a/app/lib/core/services/database/database_service.dart
+++ b/app/lib/core/services/database/database_service.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:app/core/models/animal_transport_record.dart';
 import 'package:app/core/models/transporter.dart';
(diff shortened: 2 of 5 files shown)


holtzmak (changes requested):
Can you also add the function(s) to retrieve the image from storage? I can then use these in the editing screen and so on.

holtzmak:
I watched some of that tutorial you linked. We should look into image compression, too.

MummenRider:
Will do in my next commit
> Can you also add the function(s) to retrieve the image from storage? I can then use these in the editing screen and so on.

holtzmak (approved):

MummenRider:
As discussed in the meeting, we don't need a function to retrieve image from firebase storage. We can simply use built in functionality `Image.network('imageUrl')`. Another way to do this is to use external plugin called [extended_image](https://pub.dev/packages/extended_image). This plug in have a lot features like caching, crop image and more. The query is to be used in viewModel

> Will do in my next commit
> 
> Can you also add the function(s) to retrieve the image from storage? I can then use these in the editing screen and so on.

holtzmak:
> As discussed in the meeting, we don't need a function to retrieve image from firebase storage. We can simply use built in functionality `Image.network('imageUrl')`. Another way to do this is to use external plugin called [extended_image](https://pub.dev/packages/extended_image). This plug in have a lot features like caching, crop image and more. The query is to be used in viewModel

Oh, right. I knew this and still asked. Sorry about that!

PR #226, opened 2021-03-20
Completed saving image to storage

### LINKED ISSUE #207: Add ability to change the transporter's image in the application
This work goes in hand with the tasks #202 and

- [x] Use the ImagePicker/Screen in AccountEditingScreen to pick, save, and use the chosen image as the Transporter's image

This involves work with the `DatabaseService` to save/retrieve the image. Likely you need to change the database or JSON structure of `Transporter` too.

MummenRider:
The [Permission](https://pub.dev/packages/permission_handler) setup in this commit is for iOS only. Another PR will be open for android setup.
I also added todo for sign up page because of the changes made in Transporter model. 

Note: I did not use the Image picker screen in this task. The image picker I used is inside the viewmodel itself. The tutorial I followed for this task is created by [Andy](https://www.youtube.com/watch?v=KmyhznxYQ70&t=1394s)

https://user-images.githubusercontent.com/43891568/111886949-5d1a3e00-8997-11eb-9485-8829e0b7b66b.mov

### DIFF
--- a/app/lib/core/view_models/account_edit_view_model.dart
+++ b/app/lib/core/view_models/account_edit_view_model.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:app/core/models/transporter.dart';
 import 'package:app/core/services/database/database_service.dart';
 import 'package:app/core/services/dialog_service.dart';
@@ -6,13 +8,48 @@ import 'package:app/core/services/service_locator.dart';
 import 'package:app/core/view_models/base_view_model.dart';
 import 'package:app/ui/common/view_state.dart';
 import 'package:flutter/material.dart';
+import 'package:image_picker/image_picker.dart';
+import 'package:permission_handler/permission_handler.dart';
+import 'package:uuid/uuid.dart';
 
 class AccountEditViewModel extends BaseViewModel {
   final DatabaseService _databaseService = locator<DatabaseService>();
   final NavigationService _navigationService = locator<NavigationService>();
   final DialogService _dialogService = locator<DialogService>();
 
+  final _picker = ImagePicker();
+
   Transporter _account;
+  File _selectedImage;
+  File get selectedImage => _selectedImage;
+  String _imageUrl;
+  String get imageUrl => _imageUrl;
+
+  Future<void> selectImage() async {
+    await Permission.photos.request();
+    var permissionStatus = await Permission.photos.status;
+
+    if (permissionStatus.isGranted) {
+      print('granted');
+      var tempImage = await _picker.getImage(source: ImageSource.gallery);
+
+      if (tempImage != null) {
+        _selectedImage = File(tempImage.path);
+
+        notifyListeners();
+      }
+    } else {
+      _dialogService.showDialog(
+        title: 'Access Request',
+        description: 'Permission to access gallery denied',
+      );
+    }
+  }
+
+  Future<void> setImageUrl() async {
+    _imageUrl =
+        await _databaseService.uploadAvatarImage(_selectedImage, Uuid().v4());
+  }
 
   Future<void> saveTransporterAccount({
     @required String firstName,
@@ -21,16 +58,19 @@ class AccountEditViewModel extends BaseViewModel {
     @required String userPhoneNumber,
   }) async {
     setState(ViewState.Busy);
-    var updateTransporterAccount = Transporter(
-      firstName: firstName,
-      isAdmin: _account.isAdmin,
-      lastName: lastName,
-      userEmailAddress: userEmailAddress,
-      userId: _account.userId,
-      userPhoneNumber: userPhoneNumber,
-    );
-    _databaseService.updateTransporter(updateTransporterAccount).then((_) =>
-        _dialogService
+
+    setImageUrl().then((value) {
+      var updateTransporterAccount = Transporter(
+        firstName: firstName,
+        isAdmin: _account.isAdmin,
+        lastName: lastName,
+        userEmailAddress: userEmailAddress,
+        userId: _account.userId,
+        userPhoneNumber: userPhoneNumber,
+        displayImageUrl: _imageUrl,
+      );
+      _databaseService.addTransporter(updateTransporterAccount);
+    }).then((_) => _dialogService
             .showDialog(
                 title: 'Transporter Account',
                 description: 'Changes has been saved')
--- a/app/lib/core/view_models/account_edit_view_model.dart
+++ b/app/lib/core/view_models/account_edit_view_model.dart
@@ -63,7 +63,7 @@ class AccountEditViewModel extends BaseViewModel {
         userEmailAddress: userEmailAddress,
         userId: _account.userId,
         userPhoneNumber: userPhoneNumber,
-        displayImageUrl: _imageUrl ?? _account.displayImageUrl,
+        displayImageUrl: _imageUrl,
       );
       _databaseService.updateTransporter(updateTransporterAccount);
     }).then((_) => _dialogService
--- a/app/lib/core/view_models/account_edit_view_model.dart
+++ b/app/lib/core/view_models/account_edit_view_model.dart
@@ -1,5 +1,4 @@
 import 'dart:io';
-
 import 'package:app/core/models/transporter.dart';
 import 'package:app/core/services/database/database_service.dart';
 import 'package:app/core/services/dialog_service.dart';
@@ -10,7 +9,6 @@ import 'package:app/ui/common/view_state.dart';
 import 'package:flutter/material.dart';
 import 'package:image_picker/image_picker.dart';
 import 'package:permission_handler/permission_handler.dart';
-import 'package:uuid/uuid.dart';
 
 class AccountEditViewModel extends BaseViewModel {
   final DatabaseService _databaseService = locator<DatabaseService>();
@@ -30,12 +28,10 @@ class AccountEditViewModel extends BaseViewModel {
     var permissionStatus = await Permission.photos.status;
 
     if (permissionStatus.isGranted) {
-      print('granted');
       var tempImage = await _picker.getImage(source: ImageSource.gallery);
 
       if (tempImage != null) {
         _selectedImage = File(tempImage.path);
-
         notifyListeners();
       }
     } else {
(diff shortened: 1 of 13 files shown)


mansiPaTeLL (approved):
Thanks, @MummenRider. this looks awesome.

app/lib/core/view_models/account_edit_view_model.dart
```diff
@@ -6,13 +8,48 @@ import 'package:app/core/services/service_locator.dart';
...
+  final _picker = ImagePicker();
+
   Transporter _account;
+  File _selectedImage;
+  File get selectedImage => _selectedImage;
+  String _imageUrl;
+  String get imageUrl => _imageUrl;
+
+  Future<void> selectImage() async {
+    await Permission.photos.request();
```
holtzmak, line 27: I thought we already had permissions covered for `ImageScreen` @mansiPaTeLL did `ImagePicker` not need this?

You could use `ImageScreen` still here, the screen can return the image picked through [`Navigator.pushNamed`](https://flutter.dev/docs/cookbook/navigation/returning-data)

app/lib/core/view_models/account_edit_view_model.dart
```diff
@@ -6,13 +8,48 @@ import 'package:app/core/services/service_locator.dart';
...
+      _dialogService.showDialog(
+        title: 'Access Request',
+        description: 'Permission to access gallery denied',
+      );
+    }
+  }
+
+  Future<void> setImageUrl() async {
+    _imageUrl =
+        await _databaseService.uploadAvatarImage(_selectedImage, Uuid().v4());
```
holtzmak: - [x] We should not use Uuid here. There is a great possibility of two people uploading different images with same name, and we also want to avoid duplicate images if possible

I was working on this, and figure something like the following would work:
1. The filename should be the string encoding of the image (prevent exact dupes of the image):
```
base64Encode(imageFile.readAsBytesSync())
```
2. We should check if the image doesn't already exist, and if it does, just use it's URL:
```
  @override
  Future<String> getAtrImage(String fileName) async => FirebaseStorage.instance
      .ref()
      .child('atrImages/$fileName')
      .getDownloadURL();
```
3. An easier option for you though, is to just name the avatar image after the transporter's ID, because the ID is always unique! We would need to allow file overwrites in the storage rules.

app/lib/ui/views/account/account_editing_screen.dart
```diff
@@ -69,6 +69,38 @@ class _AccountEditingScreenState extends State<AccountEditingScreen> {
...
+                                )
+                              : Container(
+                                  decoration: BoxDecoration(
+                                    color: Colors.grey[300],
+                                    borderRadius: BorderRadius.circular(50),
+                                  ),
+                                  width: 100,
+                                  height: 100,
+                                  child: Icon(
+                                    Icons.camera_alt,
```
holtzmak: - [x] When I see the camera Icon, it makes me wanna tap it. You should add an `onPressed` here to pick the image too, like the button below

app/lib/core/view_models/home_screen_view_model.dart
```diff
@@ -19,6 +23,45 @@ class HomeScreenViewModel extends BaseViewModel {
...
+        notifyListeners();
+      });
+      _currentUserSubscription = _authenticationService.currentUserChanges
+          .listen((Optional<User> user) {
+        // User has logged out or is no longer authenticated, lock the ViewModel
+        if (!user.isPresent()) _cancelTransporterSubscription();
+      });
+    } else {
+      _dialogService.showDialog(
+        title: 'Launching the account screen failed',
```
holtzmak: - [x] Oops! Looks like this message was reused. Maybe call it `Launching the home screen failed`?

app/lib/ui/views/home_screen.dart
```diff
@@ -41,12 +48,29 @@ class HomeScreen extends StatelessWidget {
...
+                      child: model.transporter != null
+                          ? CircleAvatar(
+                              radius: 52,
+                              backgroundColor: NavyBlue,
+                              child: ClipRRect(
+                                borderRadius: BorderRadius.circular(50),
+                                child: Container(
+                                  color: Colors.grey[300],
+                                  child: Image.network(
+                                    model.transporter.displayImageUrl,
```
holtzmak: - [x] This will fail maybe if the image URL is null, put something else here when the transporter does not have a URL

app/lib/ui/views/account/account_editing_screen.dart
```diff
@@ -129,26 +161,45 @@ class _AccountEditingScreenState extends State<AccountEditingScreen> {
...
+                            top: 20.0,
+                          ),
+                        ),
+                        SizedBox(
+                          height: 42,
+                          width: 200,
+                          child: RaisedButton(
+                            onPressed: () {
+                              if (widget.formKey.currentState.validate()) {
+                                model.saveTransporterAccount(
```
holtzmak, line 190: - [x] Missing `displayImageUrl` when saving here

app/lib/ui/views/account/account_editing_screen.dart
```diff
@@ -69,6 +69,38 @@ class _AccountEditingScreenState extends State<AccountEditingScreen> {
...
+                          ),
+                        ),
+                        CircleAvatar(
+                          radius: 53,
+                          backgroundColor: NavyBlue,
+                          child: model.selectedImage != null
+                              ? ClipRRect(
+                                  borderRadius: BorderRadius.circular(50),
+                                  child: Image.file(
+                                    model.selectedImage,
```
holtzmak: - [x] This image could be null, you need to check this and put something else when it is

holtzmak (changes requested):
Some changes are needed in null-checks and duplicate images. I offered some solutions to duplicate images you can use.

app/lib/ui/views/home_screen.dart
holtzmak: Wait a moment, is `Image.network(null)` valid? Maybe this already works fine.

holtzmak: Nope. It must not be null:
```
  /// Creates an object that fetches the image at the given URL.
  ///
  /// The arguments [url] and [scale] must not be null.
```

app/lib/ui/views/account/account_editing_screen.dart
MummenRider: The image url is not being added in the UI screen, but instead in viewModel. The path of the selected image would be in the business logic. Please refer to the tutorial I followed

holtzmak: Hm. OK. It's code-smelly, but yes that's how it works.

app/lib/ui/views/home_screen.dart
MummenRider: If transporter have a null displayImageUrl, then that means user did not upload his/her picture (which is fine).

To solve this issue, I added a default image inside assets and will be used if displayImageUrl is empty.

app/lib/core/view_models/home_screen_view_model.dart
MummenRider: Yea, forgot to change it. Thank you!

app/lib/ui/views/account/account_editing_screen.dart
MummenRider: Approach to solve this problem: I'll use gesture detector here.

app/lib/core/view_models/account_edit_view_model.dart
MummenRider: I checked and confirmed that Image screen did not use [Permission handler plugin](https://pub.dev/packages/permission_handler).

app/lib/core/view_models/account_edit_view_model.dart
MummenRider: I like the sound of number 3, this make sense that display profile image should be 1 only. This works in my use case.

app/lib/core/view_models/account_edit_view_model.dart
holtzmak: How the heck was image picker working before if it didn't have permissions?

MummenRider: It looks like in android, you don't need to add anything for permission to work. In ios, however, I had to configure something in the info.plist.

app/lib/core/view_models/account_edit_view_model.dart
```diff
@@ -21,16 +58,19 @@ class AccountEditViewModel extends BaseViewModel {
...
+      var updateTransporterAccount = Transporter(
+        firstName: firstName,
+        isAdmin: _account.isAdmin,
+        lastName: lastName,
+        userEmailAddress: userEmailAddress,
+        userId: _account.userId,
+        userPhoneNumber: userPhoneNumber,
+        displayImageUrl: _imageUrl,
+      );
+      _databaseService.addTransporter(updateTransporterAccount);
```
holtzmak: - [x] Shouldn't this be `updateTransporter`?

MummenRider:
As per @holtzmak 's suggestion, duplication of avatar image is resolved. Please watch the video below


https://user-images.githubusercontent.com/43891568/111891043-01f74400-89b5-11eb-8616-127172dc755f.mov

Another change: If user didn't upload his/her image or if avatar failed to load, then the default image will be used. Please watch the video below.


https://user-images.githubusercontent.com/43891568/111891064-3f5bd180-89b5-11eb-9f9b-e5e84ebeb06c.mov

app/lib/humane_transport_app.dart
```diff
@@ -12,6 +12,7 @@ class HumaneTransportApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      debugShowCheckedModeBanner: false,
```
holtzmak, line 15: Does this remove the little red "DEBUG" tag in the corner of the screen?

holtzmak (approved):

app/lib/humane_transport_app.dart
MummenRider: yes

app/lib/core/view_models/account_edit_view_model.dart
mansiPaTeLL: @holtzmak @MummenRider you need to configure in iOS and I think in android as well. You can check my commits in the experimental task.

Issue #202, opened 2021-03-10
Bring back ImagePicker/Screen


holtzmak:
This screen should be able to be launched from AccountEditing and ATREditing screens.

- [x] Ensure the image picker/camera gallery picker is called when editing and choosing to add or change a photo.
- [x] Ensure the file or filename is indeed passed on after being chosen

This work goes in hand with the tasks #215 and #199 and #207

PR #227, opened 2021-03-21
Add images to editing screen

### LINKED ISSUE #199: Update ATREditing Screen to save and show acknowledgement images
Related to #215 and #202 .

- [x] Update the editing screen so the Transporter can use the image picker/screen to pick acknowledgement images, save them, and display them
- [x] You should update the `Acknowledgement` class so as to not use null and so on.

holtzmak:
Closes #199.

In this PR I have added:
1. The ability to pick images for acknowledgment
> Made some minor changes to ImageScreen to get this to work, I think ImageScreen still needs a lot of work as "picking" the image isn't working. I'll add a video below. Related to #202 .
2. Added a custom image form field to `AckInfoFormField` (not great, kinda hacky like `Form` is, but works well)
3. Updated the `ActiveScreenViewModel` with functions to check if image already exists and upload that image if it does not (or use the image if it does)
4. Made a custom way of generating image file names based on image contents (prevents duplicates)
5. Added a way to display the images chosen in the `editing screen` and in the `display screen`s
5. Updated unit tests accordingly
6. Added a mock object for those `Image.network` requests

> Maybe we should constrain the images to fit the screen better. Wonder what a good size would be since ack images could be any size

***
Changes look like:

https://user-images.githubusercontent.com/32527219/111895043-7f32b100-89d5-11eb-87a6-562e8899e83f.mp4

https://user-images.githubusercontent.com/32527219/111895045-835ece80-89d5-11eb-8d7f-f84c5512d686.mp4


MummenRider (approved):
Good job! LGTM

PR #228, opened 2021-03-21
Will pass empty string if display url is null

MummenRider:
Display image url field is still optional, if user decided not to upload image, then we'll pass an empty string. Ass shown on the video, the user did not upload anything and therefore display url field will be set to empty string

https://user-images.githubusercontent.com/43891568/111911275-820bc100-8a2a-11eb-9f2d-80c1ab79b157.mov


holtzmak (approved):
We really should have the `Transporter` constructor set it as an empty string vs. not at all (null) sometime too.

Issue #208, opened 2021-03-10
Add the SettingsScreen and AboutScreen


holtzmak:
- [x] Add the SettingsScreen like the [GUI Prototype](https://xd.adobe.com/view/9a6f1910-d3ac-4fdf-b051-c2c78510ba32-cf3c/screen/58d55b67-910a-4061-b9ab-fefba271e0b8)
- [x] We don't need dark mode
- [x] We do need an `AboutScreen` that lists the license for the app, and the licenses of software we used in the app. 

Consider these discussions as your starting point to making the Licenses screen:
1. https://stackoverflow.com/questions/62685417/how-do-i-set-up-the-showlicensepage-function-to-work-with-flutter
2. https://www.reddit.com/r/FlutterDev/comments/echhz5/how_to_list_the_licenses_from_all_3rd_party/
3. https://api.flutter.dev/flutter/foundation/LicenseRegistry-class.html

mansiPaTeLL:
I added SettingScreen and here how it looks. 

![image](https://user-images.githubusercontent.com/44125953/111950107-74e0e780-8aa7-11eb-9f51-8a37ede786a3.png)

PR #229, opened 2021-03-22
added settings screen

mansiPaTeLL:
Part of #208.

@holtzmak @MummenRider @SanaKhan1997 Please review the <SettingScreen>. This can be accessed from every page from now. 

image


holtzmak:
@mansiPaTeLL Do the other screens still have the <- button in the app bar?

holtzmak (changes requested):
You appear to have this in the Active, History, and Home screens but not ALL screens. 
I think it would suffice to have this just on the Home Screen.

- [x] In addition, could you add the <- back button into the app bar? 
> Then it would follow the pattern we have for other screens that don't have a bottom nav bar (we can here because there's just one item).

- [x] Could you space the buttons like the Home Screen? The two close buttons seem strange.

mansiPaTeLL:
> You appear to have this in the Active, History, and Home screens but not ALL screens.
> I think it would suffice to have this just on the Home Screen.

@holtzmak I think We should keep on Active, History, and Home screen. We might not need it on Editing Screens.  I have added it on Account Screen though. We also not need it on Authentication pages obviously.

mansiPaTeLL:
@holtzmak Please review the requested changes.

holtzmak (approved):
LGTM. Thanks!

PR #232, opened 2021-03-24
Add set as default ATR button

### LINKED ISSUE #211: Add the ability to set a "default" ATR that can be used when starting a new one
- [x] Add a button to `ATRPreviewCard` saying "Set as Default?" that will allow a Transporter to reuse some of that information when starting a new ATR
- [x] Decide what information is best kept between default ATRs (maybe not contingency events or feedwater rest events, for example)
- [x] Test that the default ATR stays while the user is logged in, and is removed when they log out

holtzmak:
Closes #211 .

In this PR I have:
1. Added a button to set your default ATR. This can be clicked again if misclicked.
2. Made the `SharedPreferencesService` for shared preferences use
> About shared preferences... the data exists until the app is deleted or cache cleared (when updating). So, on sign out we have to clear it manually.
3. I added notifications to use the default ATR because people might forget
4. I updated unit tests accordingly

The information kept for the default ATR does not include:
1. FWR info
2. Loading information
3. Contingency event information

**These are different between transports, so the transporter should input new information each time.**

***

Changes look like:

https://user-images.githubusercontent.com/32527219/112397238-4f322900-8cc7-11eb-8514-20455914b4ed.mp4

https://user-images.githubusercontent.com/32527219/112397243-535e4680-8cc7-11eb-8c49-3d3d754fecad.mp4

### DIFF
--- a/app/lib/core/models/acknowledgement_info.dart
+++ b/app/lib/core/models/acknowledgement_info.dart
@@ -12,7 +12,7 @@ class AcknowledgementInfo {
       @required this.transporterAck,
       @required this.receiverAck});
 
-  factory AcknowledgementInfo.defaultAckInfo() =>
+  factory AcknowledgementInfo.empty() =>
       AcknowledgementInfo(shipperAck: "", transporterAck: "", receiverAck: "");
 
   @override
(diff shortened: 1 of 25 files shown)


MummenRider (approved):
LGTM

PR #233, opened 2021-03-25
set the state to idle when system catches error

### LINKED ISSUE #210: Test the administration POV of the application
- [x] Make an Admin account that we can reuse in demo, etc.
- [x] Test that when the admin is logged into the application, they can't use the `New` or `Active` buttons. An admin is not a Transporter.

MummenRider:
This commit shows that admin is unable to create new form both in home and active screen

admin

admin2


holtzmak (approved):
Great! Thanks for finding this.

PR #236, opened 2021-03-27
Email screen Updated

### LINKED ISSUE #201: Bring back EmailScreen
Technically, the email screen is already in the application but it needs some work:

- [x] Refactor EmailScreen to be easier to read (extract functions, use better names, etc.)
- [x] Update EmailScreen GUI to be more up-to-date with the rest of the application
- [x] Make a note on the screen that clicking the "send" button it will open your default mail application 
- [x] Separate EmailScreen from PDFScreen. 

Right now we are using navigation wrong with EmailScreen/PDFScreeen.
```
  onPressed: () {
                            Navigator.pop(context); // Don't push and pop! Just push!
                            Navigator.push( // Should also use NavigationService
                                context,
                                MaterialPageRoute(
                                    builder: (context) => EmailScreen(
                                          pdf: snapshot.requireData,
                                        )));
                          },
```

SanaKhan1997:
Capture1
Capture

### DIFF
--- a/app/lib/ui/widgets/utility/email_form.dart
+++ b/app/lib/ui/widgets/utility/email_form.dart
@@ -1,5 +1,5 @@
-import 'package:app/core/services/validation_service.dart';
 import 'package:app/core/services/service_locator.dart';
+import 'package:app/core/services/validation_service.dart';
 import 'package:app/ui/common/style.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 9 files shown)


app/lib/ui/widgets/utility/email_form.dart
```diff
@@ -98,25 +109,29 @@ class _EmailForm extends State<EmailForm> {
                       border: Border.all(color: DarkerBeige),
                       borderRadius: BorderRadius.all(Radius.circular(5.0))),
                   child: Text(
-                    'By clicking send a composed email will open in your default email app',
+                    'Note: Send button below will open Gmail',
```
holtzmak: - [x] Gmail may not be present on iPhones that use the iOS mailer, this message would be better as:
"Clicking send will open your device's email application"

holtzmak (changes requested):

app/lib/ui/widgets/utility/email_form.dart
SanaKhan1997: Done

holtzmak (approved):

PR #245, opened 2021-03-28
Time limit calls over the internet so to warn Transporter when device is offline

### LINKED ISSUE #234: Warn transporters if the application cannot connect to the internet
Right now, a lot of `future`s in the application don't complete when there's no internet. We can resolve this in a few ways:

1. Time limit every future (no)
2. Have a connection watchdog that locks app when no internet (not great)
3. Figure out what uses the internet, and see if there's any patterns we can use that allows us to warn if no internet and keep app running.

One of Firestore's promises is the cache. How can we use it? I think maybe Firebase Auth likes to connect to the internet too. We should force cache when no internet.

EDIT: I think we are stuck when trying to login or sign up when no internet. We might not have this problem if the transporter is already logged in, perhaps a warning at these particular times is all that is needed here.

holtzmak:
Closes #234 .

This PR adds:
1. Time limit of 10s to internet futures as per https://stackoverflow.com/a/61518327
2. Continue using futures against the suggestion of https://stackoverflow.com/a/53551036
3. Adds `BusyOverlay` to home screen

**We should continue to use futures because they provide us with useful error messages, like permission denied. Without them, debugging would be a lot harder and more confusing for the Transporter** 

***

Changes aren't much, but when the device is offline we will see:

Blank images when offline
Blank images when offline

PDF View never completes because it can't get images from the internet, but the user can just back out from the screen rather than wait now
PDF View sits forever, but that's fine because we can leave

We see this message when a timeout occurs:
Good future error

### DIFF
--- a/app/lib/core/services/auth_service.dart
+++ b/app/lib/core/services/auth_service.dart
@@ -6,6 +6,7 @@ import 'package:app/core/services/service_locator.dart';
 import 'package:app/core/utilities/optional.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 
 /// A service wrapping FirebaseAuth
 /// Sign in persistence is guaranteed default as per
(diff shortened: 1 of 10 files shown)


MummenRider (approved):
LGTM!

PR #248, opened 2021-03-31
added about licenses page

### LINKED ISSUE #208: Add the SettingsScreen and AboutScreen

mansiPaTeLL:
@holtzmak Please review the Licenses Page. 

https://user-images.githubusercontent.com/44125953/113101214-b06f6600-91b9-11eb-8eee-52aeee5a8325.mov

### DIFF
--- a/app/lib/ui/views/setting_screen.dart
+++ b/app/lib/ui/views/setting_screen.dart
@@ -76,10 +76,10 @@ class _SettingScreenState extends State<SettingScreen> {
                                     width: 60, height: 60)),
                             applicationVersion: 'Version 0.0.1',
                             applicationLegalese:
-                                'Copyright (C) 2020 Clark Inocalla, Kelly Holtzman, Mansi Patel, Sana Khan. '
-                                'This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. '
-                                'This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. '
-                                'You should have received a copy of the GNU General Public License along with this program. If not, see https://www.gnu.org/licenses/.',
+                                'Copyright (C) 2020 Clark Inocalla, Kelly Holtzman, Mansi Patel, Sana Khan.'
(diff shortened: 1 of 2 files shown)


app/lib/ui/views/setting_screen.dart
```diff
@@ -67,7 +67,20 @@ class _SettingScreenState extends State<SettingScreen> {
...
+                          onPressed: () => showLicensePage(
+                            context: context,
+                            applicationName: "Humane Transport",
+                            applicationIcon: Padding(
+                                padding: EdgeInsets.all(9),
+                                child: Image.asset('assets/splash_screen.png',
+                                    width: 60, height: 60)),
+                            applicationVersion: 'Version 0.0.1',
+                            applicationLegalese:
+                                'Copyright (C) 2020 Clark Inocalla, Kelly Holtzman, Mansi Patel, Sana Khan. '
```
holtzmak: Looks very good! I would put more space in the paragraph to look a bit nicer.

- [ ] Make the `applicationLegalese` look nicer to read

holtzmak (changes requested):

app/lib/ui/views/setting_screen.dart
mansiPaTeLL: @holtzmak I have looked into this. Because this is their in-built page, so we kind of not able to modify this. We ca just pass the string into the function.

holtzmak: Right, we can break the string using inline `\n` though! Could you give that a try?

mansiPaTeLL: @holtzmak Is this okay? This is all I can do because there is no custom page option available for this page at this point. 
image

holtzmak: Yeah that's great!

holtzmak (approved):

PR #249, opened 2021-03-31
Implemented delete account

### LINKED ISSUE #213: Add ability to delete Transporter account in AccountEditingScreen
The functionality for this is already available in the database and database service.

- [x] Add button in AccountEditingScreen to delete the account if the Transporter wants to
- [x] This will NOT delete the Transporter's ATRs
- [x] Launch a dialog to confirm that the user really does want to delete their account

MummenRider:
Screen Shot 2021-04-02 at 12 23 48 PM

### DIFF
--- a/app/lib/core/view_models/account_edit_view_model.dart
+++ b/app/lib/core/view_models/account_edit_view_model.dart
@@ -105,7 +105,7 @@ class AccountEditViewModel extends BaseViewModel {
       cancelText: 'No',
     );
 
-    if (!dialogResponse.confirmed) return null;
+    if (!dialogResponse.confirmed) return;
     setState(ViewState.Busy);
     try {
       _navigationService.navigateTo(WelcomeScreen.route);
--- a/app/lib/core/view_models/account_edit_view_model.dart
+++ b/app/lib/core/view_models/account_edit_view_model.dart
@@ -1,12 +1,14 @@
 import 'dart:io';
 
 import 'package:app/core/models/transporter.dart';
+import 'package:app/core/services/auth_service.dart';
 import 'package:app/core/services/database/database_service.dart';
 import 'package:app/core/services/dialog_service.dart';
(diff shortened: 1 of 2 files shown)


app/lib/core/view_models/account_edit_view_model.dart
```diff
@@ -89,4 +93,29 @@ class AccountEditViewModel extends BaseViewModel {
   }
 
   void navigateToAccount() => _navigationService.pop();
+
+  Future<void> removeTransporter(String userId) async =>
```
holtzmak: - [x] Is this used even?

app/lib/core/view_models/account_edit_view_model.dart
```diff
@@ -89,4 +93,29 @@ class AccountEditViewModel extends BaseViewModel {
...
+      description: 'Do you really want to deactivate your account??',
+      confirmationText: 'Yes',
+      cancelText: 'No',
+    );
+
+    if (!dialogResponse.confirmed) return;
+    setState(ViewState.Busy);
+    try {
+      _navigationService.navigateTo(WelcomeScreen.route);
+      await _authenticationService.deleteAccount();
```
holtzmak, line 109: Does the navigation back to welcome screen still occur when deleting the account fails? (You can test this by disconnecting from the internet before hitting the deactivate button)

holtzmak (changes requested):
Could you take a screenshot of the button lol?

app/lib/core/view_models/account_edit_view_model.dart
MummenRider: Yes, it goes to welcome screen.

holtzmak: Hm, I would prefer it not to back out if failure, but not really important to fix now.

holtzmak (approved):

PR #255, opened 2021-04-06
Add progress bar

### LINKED ISSUE #204: Experiment with a progress or timeline bar on the ATREditingScreen
Experiment with a progress bar on the editing screen. It is somewhat plain right now and could use a little GUI update love. Editing the editing screen will be done over a few tasks including this one.

- [x] Consider using a plugin or standard widget like [timeline_tile](https://pub.dev/packages/timeline_tile).

**The progress should be tied to validation, because the fields are non-null we can't just say that the Transporter is 50% done because 50% of the ATR is null.**

It would be a better idea to go by whether the `ShipperInfo` is valid, because then it must be done. Therefore, they can see on the timeline that the first box or circle is filled in while the rest are empty or greyed-out (for example).

holtzmak:
Closes #204 

This PR adds a progress bar/timeline to the editing screen. I've also updated the submission message when it fails because some forms aren't done.

> With this, it's safe to see where patterns have emerged in the editing screen and some minor cleanup could be done readily in the future. I prefer waiting for patterns to develop before doing refactoring, and now is a good time for that (for future developers... we're done here lol)

***

Changes look like:

(Also includes the display screen, it shows the numbers but in simple navy blue)

https://user-images.githubusercontent.com/32527219/113668776-fa4fc480-966f-11eb-9927-f5dc029828ac.mp4

https://user-images.githubusercontent.com/32527219/113668780-fcb21e80-966f-11eb-8c26-bfc0453c0602.mp4


MummenRider (approved):
LGTM!

PR #262, opened 2021-04-09
Update validation message on editing screen to provide more guidance

holtzmak:
This PR updates validation to block submission when nested form fields are invalid. In watching Clark's instruction videos, I didn't like the error messages and such he was given (stupid nested forms) so this work fixes that some more.

@MummenRider please try this update on your emulator and verify it works for you as you expect.


MummenRider (approved):
I, Clark Inocalla, thank you for this fix.

